PR #4, opened 2021-10-19
Fix for README

mindflayer:


kidrecursive:
Heya, thanks for the PR! The typo is actually in the method invocation and not the import, must have missed this when updating for the new client module. Going to fix it elsewhere.

PR #5, opened 2021-11-13
Fix setup.py - add endpoints

jimdickinson:
`pip install astrapy` results in a broken install without this, for me.


kidrecursive:
Thanks @jimdickinson!

PR #14, opened 2023-10-17
First pass at cleanup of Astra interfaces and general restructure

erichare:
WIP but more closely matches the interface design previously discussed


erichare:
CC @clun

astrapy/serverless.py
```diff
@@ -179,11 +177,12 @@ def collection(self, collection_name):
...
 
     def get_collections(self):
-        return self.astra_client.request(
+        res = self.astra_client.request(
             method=http_methods.POST,
-            path=f"{self.namespace_path}",
+            path=f"{self.base_path}",
             json_data={"findCollections": {}},
         )
+        return res
```
rsmith49, line 185: nit: this may be left over from debugging

astrapy/vector.py
```diff
@@ -0,0 +1,242 @@
...
+        response = self.astra_client.request(
+            method=http_methods.POST,
+            path=self.base_path,
+            json_data=json_query,
+        )
+        return response
+
+    def push(self, filter, update, options):
+        json_query = {
+            "findOneAndUpdate": {"filter": filter, "update": update, "options": options}
```
rsmith49, line 86: Why do pop and push do the same thing?

astrapy/serverless.py
```diff
@@ -114,62 +151,23 @@ def delete_subdocument(self, id, subdoc):
...
-        json_data = {
-            "findOneAndUpdate": {
-                "filter": {"_id": id},
-                "update": update,
-                "options": options,
-            }
-        }
-        res = self.astra_client.request(
+    def insert_many(self, documents=None, id_path=""):
+        return self.astra_client.request(
```
rsmith49, line 159: Doesn't look like we have any tests for `documents=None` - should we either add some or make it a required param?

rsmith49 (commented):
Nice changes! Mostly had some small comments, only higher level comment would be do we want to make a base class that contains a lot of the individual `_get`, `_post`, `find_one`, etc. functionality, since there seems to be a lot of code duplication in those methods.

Overall looks great and very pythonic to me though!

astrapy/ops.py
```diff
@@ -12,36 +12,45 @@
...
-        self.client = copy.deepcopy(client)
-        self.client.base_url = DEFAULT_HOST
-        self.client.auth_header = "Authorization"
-        self.client.astra_application_token = (
-            f"Bearer {self.client.astra_application_token}"
+    def __init__(self, token):
+        self.token = "Bearer " + token
+        self.base_url = f"{DEFAULT_HOST}"
+
+    def ops_request(self, method, path, options=None, json_data=None):
```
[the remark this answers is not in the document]
erichare: Great point. Updating.

astrapy/ops.py
```diff
@@ -12,36 +12,45 @@
...
 class AstraOps:
-    def __init__(self, client=None):
-        self.client = copy.deepcopy(client)
-        self.client.base_url = DEFAULT_HOST
-        self.client.auth_header = "Authorization"
-        self.client.astra_application_token = (
-            f"Bearer {self.client.astra_application_token}"
+    def __init__(self, token):
+        self.token = "Bearer " + token
+        self.base_url = f"{DEFAULT_HOST}"
```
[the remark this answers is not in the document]
erichare: I like the second idea.... i think we always use it, i'll update!

astrapy/ops.py
```diff
@@ -12,36 +12,45 @@
...
     def get_databases(self, options=None):
-        options = {} if options is None else options
-        return self.client.request(
-            method=http_methods.GET, path=f"{PATH_PREFIX}/databases", url_params=options
+        return self.ops_request(
+            method=http_methods.GET, path=f"{PATH_PREFIX}/databases", options=options
         )
 
     def create_database(self, database_definition=None):
         r = requests.request(
```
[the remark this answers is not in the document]
erichare: To be honest, only because it was special cased previously, but i've updated it now!

tests/astrapy/test_collections.py
```diff
@@ -191,33 +191,33 @@ def test_create_subdocument(test_collection, cliff_uuid):
 
     document = test_collection.find_one(filter={"_id": cliff_uuid})
     print("SUBDOC", document)
-    assert document["document"]["addresses"] is not None
+    assert document["data"]["document"]["addresses"] is not None
```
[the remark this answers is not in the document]
erichare: It's a good point. As far as i can tell this was always the format, and the test itself just changed (it used to automatically pull the data attribute) BUT i'm going to leave this open and clarify that

astrapy/serverless.py
erichare: Considering i think the behavior of inserting with documents=None is a little confusing, i think i'll make it required for now!

astrapy/utils.py
```diff
@@ -31,9 +32,13 @@ def make_request(
...
+
     try:
-        return r.json()
+        if return_type == "json":
+            return r.json()
+        else:
+            return r
     except Exception as e:
-        logger.error(e)
+        logger.info(e)
```
rsmith49, line 42: Maybe `warning`? Especially since we just return `None` with no record of the error

astrapy/collections.py
```diff
@@ -12,34 +12,62 @@
...
+            self,
+            collection,
+            astra_db=None,
+            db_id=None,
+            token=None,
+            db_region=None
+        ):
+        if astra_db is not None:
+            self.astra_db = astra_db
+        else:
```
rsmith49, line 38: nit: Can change this to

```
if astra_db is None:
    ...
    astra_db = AstraDB(...)

self.astra_db = astra_db
...
```

rsmith49 (commented):
Looks way cleaner to me! Is there anything we lose by removing the vector and serverless clients? I imagine we'll want service-specific methods at some point, but until then probably not worth separating

erichare:
> Looks way cleaner to me! Is there anything we lose by removing the vector and serverless clients? I imagine we'll want service-specific methods at some point, but until then probably not worth separating

I had the same question @rsmith49 - i can tell you, we're not losing anything on the vector side. This was just a naming issue between the two branches (collections.py implements the same functionality). Serverless, that's where i'm not sure. It was essentially identical to collections.py but without the vector capabilities.... if its something we are supporting asap, i think we could do something with subclasses so we don't have to re-write all the functionality in two separate files that are nearly identical. But that file had actually been removed already - i had added it back blindly from an older version, but it was never included in the tests or in the usage scripts so i really do want to get clarity on if we need it

astrapy/collections.py
erichare: Done!

astrapy/utils.py
erichare: Agreed.

PR #37, opened 2023-10-24
Add single-doc upsert as a feature in the SDK

erichare:
Adds an upsert routine

### DIFF
--- a/tests/astrapy/test_collections.py
+++ b/tests/astrapy/test_collections.py
@@ -355,6 +355,41 @@ def test_find_one_document(test_collection):
     assert document["data"]["document"] == None
 
 
+@pytest.mark.describe("upsert a document")
+def test_upsert_document(test_collection, cliff_uuid):
+    new_uuid = str(uuid.uuid4())
+
+    test_collection.upsert(
+        {
+            "_id": new_uuid,
+            "addresses": {
+                "work": {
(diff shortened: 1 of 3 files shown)


tests/astrapy/test_collections.py
```diff
@@ -355,6 +355,41 @@ def test_find_one_document(test_collection):
...
+            "addresses": {
+                "work": {
+                    "city": "Seattle",
+                    "state": "WA",
+                }
+            },
+        }
+    )
+
+    document = test_collection.find_one(filter={"_id": cliff_uuid})
```
hemidactylus, line 389: I would be extra careful here and check that the documents has featured that could _only_ come from the latest upsert. If I get it right, Cliff supposedly exists already when its upsert is called. So perhaps the upsert call on cliff should add a field or something and the assertion would check for that new field as well.

Also perhaps working on a new document would make this individual test function work even if one launches pytest skipping all others or something. (I am not sure but it might be that the order of the various test functions in a module is not under our control, so the test as is could work not as intended because the next version of pytest may reshuffle call order for what we know.

I might be wrong on this (knowing all of pytest goes beyond my powers)

erichare: @hemidactylus great point! i just pushed an update to the test, what do you think?

Issue #33, opened 2023-10-24
Pagination as a higher-level method


hemidactylus:
Building on top of the methods that are 1:1 to HTTP calls.

Would dissolve the responses and present a smooth iterable of individual Documents, lazily managing usage of the underlying calls as the iterable gets exhausted.

erichare:
Merged in and completed, thanks @hemidactylus

PR #39, opened 2023-10-25
Pagination support for "find"

hemidactylus:
Addresses #33 .

Not sure about the name of the method, I tried to make it very clear what it does.
Written as a general mechanism on top of "any" base method offering pagination (currently just "find" afaik).
Returns an _iterable_ for laziness and minimal usage of memory (caller has to call `list(...)` if they so wish.


hemidactylus:
ops. wrong target branch! closing this

PR #44, opened 2023-10-26
Sl options in insert many

hemidactylus:
In order to handle the "ordered: False" flag for fallthrough-behaviour of insert many (needed for the fastest implementation of bulk upserts in the integration), I add the options parameter to the insert_many method.

(side note: looking at the method, it seems they are not uniform in which params they support, something that might need to be fixed later)


hemidactylus:
once more doing the PR to master, Scheiße

PR #45, opened 2023-10-26
options for insert_many

hemidactylus:
In order to handle the "ordered: False" flag for fallthrough-behaviour of insert many (needed for the fastest implementation of bulk upserts in the integration), I add the options parameter to the insert_many method.

(side note: looking at the method, it seems they are not uniform in which params they support, something that might need to be fixed later)

### DIFF
--- a/astrapy/db.py
+++ b/astrapy/db.py
@@ -173,8 +173,8 @@ class AstraDBCollection:
 
         return response
 
-    def insert_many(self, documents):
-        json_query = make_payload(top_level="insertMany", documents=documents)
+    def insert_many(self, documents, options=None):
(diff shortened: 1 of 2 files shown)


erichare (approved):

prabhatjha:
> it seems the  they are not uniform in which params they support,
Can you pls list these out and share with API team? @hemidactylus

hemidactylus:
@prabhatjha  I was mostly referring to the SDK code, but yes, let me merge this asap (to unlock other development) and then we'll do a more systematic mapping starting from the API all the way to this SDK

hemidactylus:
It turns out I already opened a ticket in that sense, eheheh https://github.com/datastax/astrapy/issues/46

PR #48, opened 2023-10-26
Abstract host of api

erichare:
@hemidactylus This is based on our discussion earlier


hemidactylus:
I am afraid this solution might pose a danger moving forward. The goal is to allow for dev-environment usage, but unfortunately no replacing `apps` with `api` will solve the dev env problem. Check out the current situation:

```PROD
  ASTRA_DB_BASE_URL   apps.astra.datastax.com
  DEVOPS API           api.astra.datastax.com

DEV
  ASTRA_DB_BASE_URL   apps.astra-dev.datastax.com
  DEVOPS API           api.dev.cloud.datastax.com
```

I think a more robust choice would be to provide a happy path for prod, with a default for the devops API url, but also expose a parameter in the construction (optional, set to Null and if null falling back to the prod devops url), which the caller may want to supply. Perhaps including the protocol so one day you might be able to run with http (say on a local stargate+cassandra setup, a test environment, whatever).

Also I would recomment not to read environment variables from within the SDK code but rather have these passed as constructor inputs. WDYT?

hemidactylus:
After a discussion (Cedrick, Jeremiah J.) this is the proposed consensus:

- autofilling with the prod devops URL if the API url is the prod one, but overridable if users so wish
- no default and require user to provide devops URL if API url is not the prod one
(rationales: 1 do not confuse users with mysterious Astra-specific endpoint urls, this is OSS code, and 2 no need to expose Astra-internals more than necessary)

WDYT?

hemidactylus:
LGTM! Thank you, Eric!

PR #51, opened 2023-10-26
Add action for pytest

erichare:


erichare:
NOTE. I believe this works, however it requires the ability to set our environment variables as github secrets... I do not have access to do so @clun . Maybe sometime we can quickly set that up?

clun (approved):

PR #67, opened 2023-10-30
Add optional debugging logging for each request/response

erichare:


astrapy/utils.py
```diff
@@ -19,6 +20,19 @@ class http_methods:
 package_name = __name__.split(".")[0]
 
 
+def log_request_response(r, json_data):
+    logger.debug(f"Request URL: {r.url}")
+    logger.debug(f"Request method: {r.request.method}")
+    logger.debug(f"Request headers: {r.request.headers}")
+
+    if json_data:
+        logger.debug(f"Request payload: {json_data}")
```
hemidactylus, line 29: Proposal for the next iteration: this goes to logger.info, with the json_data stripped by what look like vectors (with a placeholder instead), and logger.debug with the full vector (?)

hemidactylus (approved):

PR #71, opened 2023-11-02
Fix #42: Update the collection naming routine

### LINKED ISSUE #42: Vector-table creation name alignment
Per the latest JSON api specs (picture),

- `"size" => "dimensions"`
- `"function"` => `"metric"`

The current names are still supported but this is a good time to adapt to the new and final naming scheme.

postman-vectors

erichare:


astrapy/db.py
```diff
@@ -348,17 +348,26 @@ def get_collections(self):
...
-            jsondata = {"name": collection_name}
+        # Initialize options if not passed
+        if not options:
+            options = {"vector": {}}
+        elif "vector" not in options:
+            options["vector"] = {}
+
+        # Now check the remaining parameters - dimension
+        if dimension and "dimension" not in options["vector"]:
+            options["vector"]["dimension"] = dimension
```
hemidactylus, line 361: nitpick: passing `options={... "vector": {"dimension": 123} ...}` AND `dimension=456` has the effect of silently discarding the 456 without warning the way I see it. Possibly dangerous?

astrapy/db.py
```diff
@@ -348,17 +348,26 @@ def get_collections(self):
...
+
+        # Check the metric parameter
+        if metric and "metric" not in options["vector"]:
```
hemidactylus, line 364: similar for passing "two metrics": the standalone-passed metric would be lost without warning, no?

astrapy/db.py
erichare: So i had this same thought, on the other hand, do we want to discard what would be in options? maybe we shouldnt allow `options` to be passed at all?

hemidactylus (approved):

PR #90, opened 2023-11-09
Fix #89 - add vector_find() method

### LINKED ISSUE #89: Add vector_find() method with clearer parameter naming
Wrapper around find() with more "pythonic" syntactic sugar.

erichare:


astrapy/db.py
```diff
@@ -76,6 +76,45 @@ def _post(self, path=None, document=None):
...
+        return sort, projection
+
+    def _post_process_find(
+        self,
+        raw_find_result,
+        include_similarity=True,
+        _key="documents",
+    ):
+        # If we only have a single result, treat it as a list
+        if type(raw_find_result["data"][_key]) == dict:
```
hemidactylus, line 102: Faulty API responses would surface to the user as KeyError here, presumably. Worth adding a check for the keys perhaps (and raise a ValueError with the jsondumps of the response) ?

astrapy/db.py
```diff
@@ -160,6 +234,36 @@ def find_one_and_replace(
 
         return response
 
+    def vector_find_one_and_replace(
+        self,
+        vector,
+        *,
```
hemidactylus, line 240: I suspect it might be argued that `replacement` come before the star, as in `vector_f_o_a_r(vec, new_obj)`. I know it contradicts what I said earlier (lol) but looking at it now it does not seem too much of a stretch

astrapy/db.py
erichare: So with the new coding logic, I'm going to resolve this for now, but please comment if you think more should be done

astrapy/db.py
erichare: I actually think this, and likewise for update, is better too...

PR #94, opened 2023-11-13
Future proof the endpoint url params

erichare:
@hemidactylus this should be fully compatible with the existing structure BUT also refactors out components enough to let us make modifications easier in the future (and eliminates the parsing issue you saw before!)

### DIFF
--- a/astrapy/db.py
+++ b/astrapy/db.py
@@ -14,13 +14,15 @@
 
 from astrapy.defaults import (
     DEFAULT_AUTH_HEADER,
+    DEFAULT_JSON_API_PATH,
+    DEFAULT_JSON_API_VERSION,
     DEFAULT_KEYSPACE_NAME,
-    DEFAULT_BASE_PATH,
 )
-from astrapy.utils import make_payload, make_request, http_methods, parse_endpoint_url
+from astrapy.utils import make_payload, make_request, http_methods
 
 import logging
 import json
+import posixpath
 
 logger = logging.getLogger(__name__)
 
@@ -44,7 +46,7 @@ class AstraDBCollection:
 
         self.astra_db = astra_db
         self.collection_name = collection_name
-        self.base_path = f"{self.astra_db.base_path}/{collection_name}"
+        self.base_path = f"{self.astra_db.base_path}/{self.collection_name}"
 
     def _request(self, *args, skip_error_check=False, **kwargs):
         response = make_request(
@@ -471,6 +473,8 @@ class AstraDB:
         self,
         token=None,
         api_endpoint=None,
+        api_version=None,
+        api_path=None,
         namespace=None,
     ):
         if token is None or api_endpoint is None:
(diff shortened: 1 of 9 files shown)


astrapy/db.py
```diff
@@ -44,7 +46,7 @@ def __init__(
```
hemidactylus, line 46: we are ok that the "direct" instantiation of AstraDBCollection (= without passing a pre-created AstraDB) does not offer the advanced params (api_path and api_version) ?
I am ok with that - it's a shortcut after all.

astrapy/db.py
```diff
@@ -482,23 +486,22 @@ def __init__(
...
+        self.api_version = api_version or DEFAULT_JSON_API_VERSION
+        self.api_path = (api_path or DEFAULT_JSON_API_PATH).rstrip("/")
 
-        # Set the namespace parameter
+        # Set the namespace
         self.namespace = namespace
 
+        # Set the Base URL for the API calls
+        self.base_url = api_endpoint.rstrip("/")
+        self.base_path = posixpath.join(
```
hemidactylus, line 501: I have trouble with this line. `base_path` is not supposed to contain the domain, right? I.e. it should start at "/api/json/v1 ...." to my understanding. Yet there is self.base_url as the first parameter of posixpath.join.

I think the reason this now works is the wrong reason. Look:
```
>>> a=AstraDB("token", "https://aaa-bbb.datastax.com")
>>> a.base_path
'/api/json/v1/default_keyspace'
>>> a.base_url
'https://aaa-bbb.datastax.com'


>>> b=AstraDB("token", "https://aaa-bbb.datastax.com", api_path="trolling/you", api_version="vLOL")
>>> b.base_path
'https://aaa-bbb.datastax.com/trolling/you/vLOL/default_keyspace'
>>> b.base_url
'https://aaa-bbb.datastax.com'
```

See? `b.base_path` mistakenly contains the whole thing from https onward. `posixpath.join` forgets the parameters before a given one, if that one starts with `/` (it's a unix file system tool after all for what I can see).

So, `posixpath.join('a', 'b', 'c')` yelds `a/b/c`, but (and this is the brittle logic that underpins the above `posixpath.joins('a', 'b', '/c', 'd')` yields just `/c/d`.

I would be more fussy in stripping slashes left and right to each string, then reinstating manually the needed ones and explicitly not using self.base_url in constructing self.base_path - if I get this part right, that is.

astrapy/defaults.py
```diff
@@ -1,10 +1,12 @@
 DEFAULT_AUTH_PATH = "/api/rest/v1/auth"
-DEFAULT_BASE_PATH = "/api/json/v1"
+
+DEFAULT_JSON_API_PATH = "/api/json"
+DEFAULT_JSON_API_VERSION = "v1"
```
hemidactylus, line 4: Yeah, perhaps adding a unit test that provides the same api_path and api_versions with of without trailing _and leading_ slash, and then checks you always get the same self.api_path and self.base_path , would be a good sanity check.

astrapy/db.py
erichare: That was kind of my feeling too!

astrapy/ops.py
```diff
@@ -13,19 +13,20 @@
...
+from astrapy.defaults import DEFAULT_DEV_OPS_API_VERSION, DEFAULT_DEV_OPS_URL
 
 import logging
 
 logger = logging.getLogger(__name__)
 
 
 class AstraDBOps:
-    def __init__(self, token, dev_ops_url=None):
+    def __init__(self, token, dev_ops_url=None, dev_ops_api_version=None):
```
hemidactylus, line 24: I would always, here as in db.py, first strip slashes from both ends of such a string, then manually put them between when joining. Probably three people will start passing `/v1`, `v1` and `/v1/` here and we want them all to be happy

tests/astrapy/test_ops.py
```diff
@@ -83,7 +83,7 @@ def test_create_database(devops_client):
 @pytest.mark.describe("should create a keyspace")
 def test_create_keyspace(devops_client):
     response = devops_client.create_keyspace(
-        keyspace="test_namespace", database=os.environ["ASTRA_DB_ID"]
+        keyspace="test_namespace", database=str(uuid.uuid4())
```
hemidactylus, line 86: I guess such a test will never work. The DB_ID must exist on the test account - a random string will not. Unless I'm missing something, this should be reverted (to trusting the test to run with the DASTRA_DB_ID variable set to an existing database ID in the tester's account)

erichare: Oh this was a mistake on my part! See, i was trying to strip out the use of ASTRA_DB_ID where it was still used - but, you're right, we actually need a test db to exist. I think i just want to clarify that this is not necessarily the ASTRA_DB_ENDPOINT_URL database.... so maybe it can be called ASTRA_DB_TEST_ID or something...

erichare:
@hemidactylus when you get a chance, one more quick set of eyes on this would be fantastic :) appreciate it.

PR #96, opened 2023-11-14
Fix httpx support for concurrency

erichare:


hemidactylus:
I think the `AstraDBOps` class and the  `make_request` invocation in `ops.py` should also get the "client treatment". right now that class is likely broken (?)

hemidactylus:
I'm not an expert on httpx at all ... so, what is the reason to make the client a _class_ attribute? (and not, at that point, a module-level gobal variable or something?)

hemidactylus:
This needs extra care to avoid accidentally breaking OpenAI ([they](https://github.com/openai/openai-python/issues/690) had a problem with httpx 0.25.1 and apparently solved it by downgrading httpx). We certainly need to test a few openAI/langchain things carefully before merging

PR #101, opened 2023-11-15
Fix for upserting of documents

erichare:


astrapy/db.py
```diff
@@ -648,11 +652,8 @@ def upsert(self, document):
                 filter={"_id": document["_id"]},
                 replacement=document,
             )
-            upserted_id = result["data"]["document"]["_id"]
-        else:
-            upserted_id = result["status"]["insertedIds"][0]
 
-        return upserted_id
+        return result
```
hemidactylus, line 656: but it seems that now the method does not return _The _id of the inserted or updated document_ anymore, as it should per docstring. I think the _id is found within the response somewhere, albeit possibly in two different places depending whether the insert or the replace succeed

erichare: My only wonder here was, since in theory there could be an error that we're covering up, do we want to assume the return indeed has the id, or do we just want to return the raw result. I know the previous behavior did that... but just curious...

PR #104, opened 2023-11-16
Explicitly use a push and pop notation

### LINKED ISSUE #16: Push and Pop have identical code

erichare:
Fix #16


hemidactylus (approved):
I don't know much on these push/pop things in the vector API but looking at the postman collection it seems all right. LGTM :)

PR #111, opened 2023-11-16
Split requirement files and automates reqs in setup.py

### LINKED ISSUE #108: Introduce "dev requirements" and tie the setup.py reqs to the main reqfile
Split into requirements.txt and requirements-dev.txt

The latter with pytest, ruff, etc

The former with the reqs we want to bundle as package deps ... _plus_ the special `-e .` to self-install in a local dev env.

*Solves* the risk if misaligning setup.py and requirements.txt (also keeping dev stuff out of the build)

hemidactylus:
split requirement files and automates reqs in setup.py

Closes #108 

Only the reqs needed for the bundle (plus `-e .`) are in `requirements.txt`. these (minus the special one) are pulled by `setup.py` to avoid drift.

Devs, for testing etc, would additionally install from `requirements-dev.txt`.

Incidentally, installing with -e . makes testing easier by just running `pytest`.

I tried several ways to install in a clean virtualenv and everything worked. In particular I tested building the wheel, installing it and importing all astrapy parts - success!


erichare (approved):
I love this!!

PR #115, opened 2023-11-21
Full support for type hints

### LINKED ISSUE #28: Full typing support

hemidactylus:
This PR adds 100% coverage of type hints to both the library and the test suite.

Closes #28 

Most of the changes in the code are "trivial" and amount to adding the predictable type annotations for function parameters and return values. However, in a few locations, additional changes were convenient or required (reasons: making typing easier to manage/fix errors made visible with typing).To make review easier, here they are:

- Removed args/kwargs in `_request` (see make_request, I favored explicit listing, also less type-y and no risk of passing same param twice (of which mypy complains). Also no point in silently accepting kwargs and never using them, it's only a risk of not noticing future problems IMO
- In the `AstraDB` class the whole `_put` was invoked ... but was missing altogether. (please check if its impl is convincing)
- the post-processing associated to the `vector_*` methods was removed: its ambiguous `List[docs]/doc` type makes things error-prone and hard to handle cleanly. The caller knows if it's list or single-doc, and now this is how the code works
- vector_find and also _pre_process_find had a "return ValueError(...)" instead of "raise ValueError(...)" by mistake (fixed)
- pagination reworked: introduced partialing of the unpaginated find + Protocol class to capture the type (with its named `options` parameter) of the only input remaining free after partialing
- `ops.py`: added a wrapper for "make this response into a json" for better typing mgmt
- in test_ops I made the `environ.get`s into `environ["..."]` (token, db ID). It makes hardly sense not to (and accordingly helps typing)
- `ops.py`, in `update_datacenter_endpoint` made `endpoint` not Optional (why it was in the first place) as it is indexed in the function body


hemidactylus:
I should also comment that
- I ran all tests (including ops.py) on python3.8 and python3.9
- I added `py.typed` and instructed `setup.py` to bundle it with the library
- I added a note on the README (section "for developers") about routine style+linter+type checks to library _and_ tests

erichare (approved):

PR #123, opened 2023-11-24
Fix bug in vector_ methods and rearrange tests

### LINKED ISSUE #122: vector_find_* breaks when no docs found
BUG: when vector_find_* finds no matches, currently, boom!

Vector_* methods always return not the raw API response, rather a list of docs/a doc-or-null.

- align docstrings
- remove the obsolete postprocess step
- enhance tests
- handle nulls in the three vector_finds that return a single item

hemidactylus:
### Original scope

This PR fixes the bug in #122 whereby in absence of any match, the `vector_*one*` methods would error with a "NoneType is not iterable".
With the introduction of usage of `includeSimilarity` in the API calls, the whole postprocessing step in all `vector_*` methods is not required and has been removed (simplification!).
Also the docstrings are adapted accondingly.

Closes #122 

### Testing

To better highlight the issue, I started adding to the tests ... then realized a major refactoring of the whole test suite was in order and I just went ahead with that. So, _this PR also contains a major improvement to the tests._

- Each test function can now be run as stand-alone (e.g. through `pytest ... -k <selector>`) in that it does not rely on any other test function to run, in any order. This warranted a `conftest.py` and a rethinking of all the collection fixtures
- To optimize DB usage, there are a few "read-only" session-wide collection fixtures, "writeable" session-wide ones (each function supposed to generate IDs to avoid collisions) and "disposable" function-wide fixtures to ensure isolation of tests.
- Also there were several missing tests, they were arranged in a non systematic fashion, named in unexpected ways, etc - to which I put a global, understandable order (see below for the new test names).
- Thanks to the fixtures above, I refactored the tests in more modules for easier management: schema stuff / pagination stuff / `vector_` methods / regular "db.py" stuff.
- And of course the ops stuff: to encourage occasional running, I added support for the `TEST_ASTRADBOPS` env var (as opposed to uncommenting a `skipif` in the code). Still not running by default.
- Also to cover the fact that the typical token by the user is not strong enough to do Ops stuff (yet is the token we want to test with), I introduce an optional "ops-specific token" overriding the "regular" token if desired.
- Cliff Wicklow is not dead, by the way, but now he's just making a cameo (blame the optimization of the collections :)
- The README now gives the basic options to manage testing (including the proper env vars, management of the optional OPS and how to suppress the extended logging).

As an aside, here are now the test function names (unchanged ops-test functions at the end):

```
test_path_handling
test_create_use_destroy_nonvector_collection
test_create_use_destroy_vector_collection
test_get_collections
test_find_paginated
test_truncate_collection_fail
test_truncate_nonvector_collection
test_truncate_vector_collection
test_find_one_filter_novector
test_find_filter_novector
test_find_find_one_projection
test_find
test_find_error
test_find_limitless
test_create_document
test_insert_many
test_insert_many_ordered_false
test_upsert_document
test_update_one_create_subdocument_novector
test_delete_subdocument_novector
test_find_one_and_update_vector
test_find_one_and_update_novector
test_find_one_and_replace_vector
test_find_one_and_replace_novector
test_delete_one_novector
test_delete_many_novector
test_pop_push_novector
test_vector_find
test_vector_find_projection
test_vector_find_filters
test_vector_find_one
test_vector_find_one_and_update
test_vector_find_one_and_replace

test_client_type
test_get_databases
test_create_database
test_create_keyspace
```

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -384,7 +384,9 @@ Ensure you provide all required environment variables (you can do so by editing 
 export ASTRA_DB_APPLICATION_TOKEN="..."
 export ASTRA_DB_API_ENDPOINT="..."
 export ASTRA_DB_KEYSPACE="..."              # Optional
+
 export ASTRA_DB_ID="..."                    # For the Ops testing only
+export ASTRA_DB_OPS_APPLICATION_TOKEN="..." # Ops-only, falls back to the other token
 ```
 
 then you can run:
--- a/README.MD
+++ b/README.MD
@@ -378,16 +378,25 @@ black --check tests && ruff tests && mypy tests
 
 ### Testing
 
-Ensure you provide all required environment variables:
+Ensure you provide all required environment variables (you can do so by editing `tests/.env` after `tests/.env.template`):
 
 ```bash
-export ASTRA_DB_ID="..."
 export ASTRA_DB_APPLICATION_TOKEN="..."
 export ASTRA_DB_API_ENDPOINT="..."
+export ASTRA_DB_KEYSPACE="..."              # Optional
+export ASTRA_DB_ID="..."                    # For the Ops testing only
 ```
 
 then you can run:
 
 ```bash
 pytest
+```
+
+To remove the noise from the logs (on by default), run `pytest -o log_cli=0`.
+
+To enable the `AstraDBOps` testing (off by default):
+
+```bash
+TEST_ASTRADBOPS=1 pytest
 ```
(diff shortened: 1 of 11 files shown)


erichare (approved):
I absolutely love the test rewrite and thank you so much for fixing those vector() bugs and updating the tests. I ran through my tests:

Running tests locally, example script, langchain, and llama integrations... all looks great.

PR #125, opened 2023-11-27
Raise HTTP errors

### LINKED ISSUE #124: Errors in a request hidden by JSON decoding error
When the HTTP request returns an error, there is no JSON body, which means that the call to `json()` will fail like this:

```
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Unfortunately, that suppresses the actual errors (in my case I was getting HTTP 413 "Request entity too large"). I had to manually patch astrapy to find that error. 

Ideally, either `make_request` would fail if it received an error (https://github.com/datastax/astrapy/blob/master/astrapy/utils.py#L86) by calling `r.raise_for_status()` to raise HTTP errors as exceptions, or anything that used `make_request` would check to make sure there was `json` to decode before calling `.json()`.

bjchambers:
This fixes #124.

### DIFF
--- a/astrapy/ops.py
+++ b/astrapy/ops.py
@@ -19,7 +19,7 @@ import httpx
 
 from astrapy.utils import make_request, http_methods
 from astrapy.defaults import DEFAULT_DEV_OPS_API_VERSION, DEFAULT_DEV_OPS_URL
-from astrapy.types import API_RESPONSE
+from astrapy.types import API_RESPONSE, OPS_API_RESPONSE
 
 
 logger = logging.getLogger(__name__)
@@ -69,7 +69,7 @@ class AstraDBOps:
         path: str,
         options: Optional[Dict[str, Any]] = None,
         json_data: Optional[Dict[str, Any]] = None,
-    ) -> API_RESPONSE:
+    ) -> OPS_API_RESPONSE:
         req_result = self._ops_request(
             method=method,
             path=path,
@@ -77,11 +77,13 @@ class AstraDBOps:
             json_data=json_data,
         )
         return cast(
-            API_RESPONSE,
+            OPS_API_RESPONSE,
             req_result.json(),
         )
 
-    def get_databases(self, options: Optional[Dict[str, Any]] = None) -> API_RESPONSE:
+    def get_databases(
+        self, options: Optional[Dict[str, Any]] = None
+    ) -> OPS_API_RESPONSE:
         """
         Retrieve a list of databases.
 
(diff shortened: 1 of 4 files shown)


hemidactylus:
First, thank you @bjchambers for spotting and improving the behaviour on errors!

I ran extensive testing with the patched code and it all LGTM - including the upsert (the reason being that "errors" such as _document already exist_ is not an HTTP error in these API, rather a legitimate returned response the caller has the responsibility to check. So we're good).

So, TLDR = LGTM.

But if you want to know more (@erichare , please do) read on, since I pushed further changes (mostly the tests) to this PR to bring a flawless test suite to master, including the Ops tests (which we seldom run and not in the CI, but still.)

Enabling the `raise_for_status` is all good and a welcome thing to do, but it uncovered some trouble in the Ops test which I improved.

- The DB creation test was silently failing in the "delete the newly-created db" part. I took that out [1]
- The create keyspace test was meaningless, since it was targeting a nonexisting DB and then not checking the obvious call failure that came with that. Fixed.
- More importantly, while the JSON API always return a top-level _dict_, the DevOps is more heterogeneous. Some calls, such as `get_databases`, return a _list_. This god in the way with typing and made me add some stuff to correctly handle that. But I just patched so that it always works, with a few `cast`s around to satisfy mypy. More work will be needed in the ops sector. [2]

[1] = on the roadmap we should have a more thorough Ops testing, which is not a trivial thing. There, I'd like to test db deletion as well, probably with a very long-running test that polls and waits until the DB is deletable (as long as it's being created you can't). And other similar operations.
[2] = In particular, I added a union type covering lists as well, adapted the docstring of the `get_databases` method, and the corresponding code. I think the right (and type-sensible way) would be to have two top-level return types, the dicty and the listy, and have two separate `_json_ops_request`, one per type. This would "document in code" what the Ops API does in the proper way.

erichare (approved):
@hemidactylus Very good catch on the ops stuff. I'm approving because, even though the proper typing etc is something we need to do, the ops client itself is already in kind of a, shall we say, less than ideal state, and i think this gets us further and raising for errors is such an important change (so the user gets more feedback about the result)

We still have the separate issue discussed with Enrico about errors that produce a successful HTTP response, but this doesn't impact it.

I say let's merge. At some point i'll bump the version to 0.6.3 and start a changelog (no plans to push to pypi of course) but just so we can eventually push a pre-release or just install from git easily and know we're on the latest version.

Thanks @bjchambers !!

PR #128, opened 2023-11-29
Fix #126 - improve handling of API errors

### LINKED ISSUE #126: ValueError is hard to use
I am using AstraPy to perform a lookup.

I am forcing the API to return an error, and I want to take some actions depending on the error.

```python
        try:
            collection.find_one(filter={
                'metadata.chunks': {'$gt': 2}
            })
            pytest.fail("Should have thrown ValueError")
        except ValueError as e:
            # This looks very ugly, but it's the only way to get the error message
            error = json.loads(e.args[0])[1]
            assert error.get("errorCode") == "UNSUPPORTED_FILTER_OPERATION"
```

Currently the ValueError object has no easy way to be manipulated, it contains a string with a JSON that contains a raw error coming from the serve.

# Proposal
Return a better representation of the errors, with some data structure that can be handled by the application

erichare:
This basically creates a wrapper class that processes API requests, handling errors in the API response more elegantly and allowing us to have one central point to manage them.


eolivelli (commented):
Is it possible to access the error code as a field of the errors? For instance DOCUMENT_ALREADY_EXISTS

erichare:
> Is it possible to access the error code as a field of the errors? For instance DOCUMENT_ALREADY_EXISTS

So currently the output is an APIRequestError like this:

```
APIRequestError: API returned an error: [{'message': "Failed to insert document with _id '1': Document already exists with the given _id", 'errorCode': 'DOCUMENT_ALREADY_EXISTS'}]
```

Is that sufficient for this, or are you saying you'd like to extract that `errorCode` field directly?

astrapy/api.py
```diff
@@ -0,0 +1,71 @@
+import httpx
+
+from typing import cast
+
+from astrapy.types import API_RESPONSE
+from astrapy.utils import make_request
+
+
+class APIRequestError(Exception):
```
hemidactylus, line 9: I would add a member that contains the very httpx.RequestError, not made into a string, to avoid swallowing useful info from clients who may want to use that stuff

astrapy/api.py
```diff
@@ -0,0 +1,71 @@
...
+    def __init__(self, message):
+        super().__init__(message)
+
+
+class APIRequestHandler:
+    def __init__(
+        self,
+        client,
```
hemidactylus, line 17: Type hints (does mypy not complain here?0

astrapy/api.py
```diff
@@ -0,0 +1,71 @@
...
+            return self._process_response(response)
+        except httpx.RequestError as e:
+            raise APIRequestError(f"An error occurred while making the request: {e}")
+        except httpx.HTTPStatusError as e:
+            raise APIRequestError(
+                f"The API returned an error while handling the request: {e}"
+            )
+
+    def _process_response(self, response: httpx.Response):
+        if (
```
hemidactylus, line 59: Needs careful checks that the DevOps api does not do weird stuff such as returning 3xx in "all right" cases (devops api has a history of being a bit weird)

astrapy/api.py
```diff
@@ -0,0 +1,71 @@
...
+            not 200 <= response.status_code < 300
+        ):  # TODO: Is this the right range? Any abstractions?
+            raise APIRequestError(
+                f"Non-success status code received: {response.status_code}"
+            )
+
+        responsebody = cast(API_RESPONSE, response.json())
+
+        if "errors" in responsebody and not self.skip_error_check:
```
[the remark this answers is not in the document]
hemidactylus: No, this is a test whether a key is in the `responsebody` associative map. No risk of accidental collisions here :)

astrapy/api.py
```diff
@@ -0,0 +1,71 @@
...
+            raise APIRequestError(f"API returned an error: {responsebody['errors']}")
```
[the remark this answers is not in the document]
hemidactylus: See my remark above (additional members for the original exception and, if present, the response in its entirety)

astrapy/api.py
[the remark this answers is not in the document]
eolivelli: Sometimes the API does not report that entry, for instance this error is not matching this pattern:

`[{'message': "Failed to insert document with _id '5eb4789401d2433182e3ecd5df6660d6': INVALID_ARGUMENT: Zero vectors cannot be indexed or queried with cosine similarity"}]`

This happens when you try to index an empty text


I am afraid that we miss some other cases.
(I am currently writing tests on RAGStack about all the most common error cases)

astrapy/api.py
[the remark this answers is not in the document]
eolivelli: Totally agreed !

eolivelli:
> @eolivelli I left my suggestion for how to expose the very original exception without swallowing too much of it.
> 
> That being said, unfortunately expecting a _single_ error code is not always the right thing. Indeed, the `errors` field (plural) is an array in the response. So I think there should be no "error code" field at all in the returned exception, possiblt a list (but of limited use at that). Here is the response from a particular insert_many to make this point:
> 
> ```
> {
>     "errors": [
>         {
>             "errorCode": "DOCUMENT_ALREADY_EXISTS",
>             "message": "Failed to insert document with _id \"b\": Document already exists with the given _id"
>         },
>         {
>             "errorCode": "DOCUMENT_ALREADY_EXISTS",
>             "message": "Failed to insert document with _id \"c\": Document already exists with the given _id"
>         }
>     ],
>     "status": {
>         "insertedIds": [
>             "40421610-e60d-4f6d-b4b0-39232cd682b4"
>         ]
>     }
> }
> ```

Saw that.
We should add a way to inspect the structure.

hemidactylus:
@erichare another general note (an implied assumption above which I'd like to quickly check).
We are going to a single exception for two distinct types of "errors while using the API" here.
- 300x, 400x, anything from the network layer
- "errors" such as: wrong vector size, doc already exists, etc etc, where the API receives the request and internally returns a 200 with an errorful response.

Are we sure this is what we need? This is effectively masking the network-layer errors and lumping them into a big undifferentiated group, is this what we want?
Why not let httpx raise its own exceptions, undisturbed, and leave it to the client to parse them - astrapy only taking care of its own specific "errors" (the second type above).

I am generally against masking errors, they should normally bubble up as untouched as possible IMO.

hemidactylus:
> Saw that. We should add a way to inspect the structure.

As in, astrapy _offers_ the whole response and then the client is free to inspect, right?
(Not: "astrapy does the inspection itself and reshuffles the stuff therein", I surmise)

eolivelli:
FYI
https://github.com/datastax/ragstack-ai/pull/63

astrapy/api.py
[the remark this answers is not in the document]
hemidactylus: Oh, no! This is a CQL-layer error (more precisely SAI stuff: null vector has no well-defined angle).

Possibly the errors that are not originating at the API layer, rather below, end up like this. I wonder if this should be checked and raised by AstraPy similarity as what now looks for "errors" member in the returned json (... but a bit less elegant to do and collision-prone, also sensitive to change of phrasing of the error message. Not very nice to do)

hemidactylus:
> The only problem with this approach is that the application will have to deal directly with httpx exception, and implicitly depend on it

I see your point. The alternative, however, is clients to import an exception from astrapy, which is not that different (a bit more unknown, perhaps, and -let's be honest- less tested than httpx). Consider most clients would not know of astrapy otherwise, as they would import LangChain/RAGStack stuff.

Let us collect opinions here. Swallowing/recasting/pruning errors can quickly become a way to make developers' life miserable, this is why I urge to caution here (the pre-vector-API python and javascript clients as big offenders here marked me for life... lol)

erichare:
So, "requirements" for this:

1. We want to use the raw errors, rather than "masking them
2. But we also want to be able to access the API-produced "errors" as a data structure for processing

So here is my proposal. Effectively we follow these steps:

1. A request is made to the API
2. If there is an error raised by the API itself, we raise that error directly, i.e., `r.raise_for_status()` as is being done starting this week.
  ```python
      def request(self: T) -> API_RESPONSE:
          # Make the raw request to the API
          self.response = self._raw_request()
  
          # If the response was not successful (non-success error code) raise an error directly
          self.response.raise_for_status()
  
          # Otherwise, process the successful response
          return self._process_response()
   ```
3. Assuming we get to `_process_response()`, it was a success code.
4. We attempt to parse the response, if we have "errors" in the response body, we return a dictionary containing the response, the status code (likely 200 for most), and an indication that the call errored, such that we can process programatically any of the errors in these cases.
5. Otherwise, if no error, we return the response body as normal
  ```python
          def _process_response(self: T) -> API_RESPONSE:
            # In case of other successful responses, parse the JSON body.
            try:
                # Cast the response to the expected type.
                response_body: API_RESPONSE = cast(API_RESPONSE, self.response.json())
        
                # If the API produced an error, warn and return the API request error class
                if "errors" in response_body and not self.skip_error_check:
                    logger.warning(response_body["errors"])
                    return APIRequestError(self.response).get()
        
                # Otherwise, set the response body
                return response_body
            except ValueError:
                # Handle cases where json() parsing fails (e.g., empty body)
                return APIRequestError(self.response).get()
 ```
 
 This im sure needs refinement, but i think it gets us a little closer to what you guys want - not masking errors, programmatically accessing the errors, and having a framework for managing this as the API evolves.

astrapy/api.py
erichare: Yes. I've now made it only raise_for_status which treats 3xx as "ok", so i think now we should be good

astrapy/api.py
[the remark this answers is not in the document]
erichare: Raw errors and the raw response are now returned in the event of an error (note, if the API itself returns a 400 / 500 etc error code then we don't get this right now - we actually raise the error directly - but and of the "success" + error responses will allow this.

astrapy/api.py
```diff
@@ -0,0 +1,91 @@
...
+        return self._process_response()
+
+    def _process_response(self: T) -> API_RESPONSE:
+        # In case of other successful responses, parse the JSON body.
+        try:
+            # Cast the response to the expected type.
+            response_body: API_RESPONSE = cast(API_RESPONSE, self.response.json())
+
+            # If the API produced an error, warn and return the API request error class
+            if "errors" in response_body and not self.skip_error_check:
```
eolivelli, line 83: as discussed elsewhere, not all the error responses contain an "errors" key,
probably we need to handle pass a function "response_contains_errors" and implement custom handling for each different API call

ideally the API should return a non 200 HTTP Code in case of errors. I know that this is not the case, so here in AstraPy we cannot do much

Maybe now that we are in "preview" we can raise the problem in Stargate and fix it upstream (cc @clun @prabhatjha )

erichare: Updated the logger, but yes.... i'm not sure exactly all the possible error conditions that can occur and how we know they're errors exactly. It feels like if possible this would be ideal to fix upstream for sure, or at least if its well documented somewhere and a good reason for what possible errors we might get and the format of said responses.

erichare:
I will respond it full Monday, but my primary question i have - i agree with basically everything you said, but how do we expect the caller to be able to catch and parse the errors as per Enrico's request? I'd much prefer to always raise an error, but that was the original behavior i had setup and the issue was the expectation that when "errors" are returned from the API, we be able to parse them and read information from them.

eolivelli (commented):
can you please add some tests cases about handling the errors ?
this will help a lot in seeing who users will deal with errors and also it will help us not breaking the behaviour in the future

erichare:
> can you please add some tests cases about handling the errors ? this will help a lot in seeing who users will deal with errors and also it will help us not breaking the behaviour in the future

Great idea. i just pushed one - we could add more, though the basic idea of how we handle it is almost always exactly like this. The test i added inserts a document, then attempts to insert a document with the same id (raises an error). Then, we attempt to insert it again, but catch the exception, and parse out the error message... which i think is the flow most people would want!

astrapy/exceptions.py
```diff
@@ -0,0 +1,83 @@
...
+class APIRequestError(ValueError):
+    def __init__(self, response: httpx.Response) -> None:
+        self.response = response
+        super().__init__(response.text)
+
+    def __repr__(self) -> str:
+        return f"{self.response}"
+
+
+class APIRequestHandler:
```
eolivelli, line 23: nit: the name of the file is "exceptions.py", but it actually contains the code that does the http class, maybe we can have a better name ?

tests/astrapy/test_db_dml.py
```diff
@@ -362,6 +365,50 @@ def test_insert_many_ordered_false(
     assert check_response["data"]["document"]["_id"] == _id1
 
 
+@pytest.mark.describe("test error handling")
+def test_error_handling(writable_vector_collection: AstraDBCollection) -> None:
+    _id1 = str(uuid.uuid4())
+
+    result1 = writable_vector_collection.insert_one(
```
eolivelli, line 372: can you please also add a test about a network error ? like unreachable server (just configure a dummy url like localhost:1234)

eolivelli (commented):
I think that we are on our way.

I left only a comment about adding another test

astrapy/exceptions.py
erichare: I agree. Maybe `api.py`? it's still not exactly the most clear, but since the request handlers and exception classes all are derived and related to making API requests...

erichare:
@eolivelli i did a renaming of that exceptions.py to api.py which i think better reflects its role in the package. I also added a test, with an invalid endpoint URL that generates an httpx ConnectError. Let me know what you think!

EDIT: having some issues with the tests, but it looks like server instability (random 500 errors)...

astrapy/api.py
```diff
@@ -0,0 +1,150 @@
...
+from astrapy.utils import amake_request, make_request
+
+T = TypeVar("T", bound="APIRequestHandler")
+AT = TypeVar("AT", bound="AsyncAPIRequestHandler")
+
+
+logger = logging.getLogger(__name__)
+
+
+class APIRequestError(ValueError):
```
cbornet, line 15: I would make `APIRequestError` depend directly on `Exception` as it has different meaning than `ValueError`
```suggestion
class APIRequestError(Exception):
```

astrapy/api.py
```diff
@@ -0,0 +1,150 @@
...
+    def __init__(self, response: httpx.Response) -> None:
+        super().__init__(response.text)
+
+        self.response = response
+
+    def __repr__(self) -> str:
+        return f"{self.response}"
+
+
+class APIRequestHandler:
```
cbornet, line 25: I'm not sure to see the interest of the `APIRequestHandler ` compared to static utils methods.

PR #132, opened 2023-11-30
Fix #130: Use path in the _post calls

### LINKED ISSUE #130: The `path` argument to `_post` is ignored in AstraDBCollection
Quick fix: use it in preparing `full_path`, as happens in the _get and _put sibling methods

(that being said, nobody uses that arg, but it does not hurt to have it)

erichare:


hemidactylus (approved):
LGTM :)

PR #134, opened 2023-12-01
First pass at batched versions of insert / upsert

erichare:


astrapy/db.py
```diff
@@ -694,6 +697,51 @@ def insert_many(
 
         return response
 
+    def batched_concurrent_insert_many(
```
cbornet, line 700: It seems that `batched` is redundant with `many`.
Maybe `concurrent_insert_many` would be enough ?

astrapy/db.py
```diff
@@ -694,6 +697,51 @@ def insert_many(
...
+            documents (list): A list of documents to insert.
+            options (dict, optional): Additional options for the insert operation.
+            partial_failures_allowed (bool, optional): Whether to allow partial failures in the batch.
+            batch_size (int, optional): Override the default insertion batch size.
+            concurrency (int, optional): The number of concurrent batch insertions.
+        Returns:
+            list: The responses from the database after the batched insert operation.
+        """
+
+        # Split the documents into batches
```
cbornet, line 720: It's more a `chunk` than a `batch` IMO.

astrapy/db.py
```diff
@@ -825,6 +873,29 @@ def upsert(self, document: API_DOC) -> str:
...
+            concurrency (int, optional): The number of concurrent batch updates.
+
+        Returns:
+            List[str]: A list of "_id"s of the inserted or updated documents.
+        """
+
+        # Perform the bulk upsert with concurrency
+        with ThreadPoolExecutor(max_workers=concurrency) as executor:
+            futures = [executor.submit(self.upsert, document) for document in documents]
+            response_list = [future.result() for future in futures]
```
cbornet, line 895: Same here, the returned value could be `List[Union[str, Exception]]`

astrapy/db.py
```diff
@@ -694,6 +697,51 @@ def insert_many(
...
+            logger.debug(f"Processing batch #{index + 1} of size {len(batch)}")
+
+            return self.insert_many(
+                documents=batch,
+                options=options,
+                partial_failures_allowed=partial_failures_allowed,
+            )
+
+        # Perform the bulk insert with concurrency
+        with ThreadPoolExecutor(max_workers=concurrency) as executor:
```
cbornet, line 736: If concurrency == 1, maybe don't use a thread-pool ?

astrapy/db.py
```diff
@@ -825,6 +873,29 @@ def upsert(self, document: API_DOC) -> str:
 
         return upserted_id
 
+    def batched_concurrent_upsert(
```
cbornet, line 876: Name it `upsert_many` ?

astrapy/db.py
```diff
@@ -694,6 +697,51 @@ def insert_many(
...
         return response
 
+    def batched_concurrent_insert_many(
+        self,
+        documents: List[API_DOC],
+        options: Optional[Dict[str, Any]] = None,
+        partial_failures_allowed: bool = False,
+        batch_size: int = MAX_INSERT_BATCH_SIZE,
+        concurrency: int = 1,
+    ) -> List[API_RESPONSE]:
```
cbornet, line 707: Would it be interesting/possible to merge the results into a single API_RESPONSE ?

astrapy/db.py
```diff
@@ -825,6 +873,29 @@ def upsert(self, document: API_DOC) -> str:
...
+            documents (List[dict]): The documents to insert or update.
+            concurrency (int, optional): The number of concurrent batch updates.
+
+        Returns:
+            List[str]: A list of "_id"s of the inserted or updated documents.
+        """
+
+        # Perform the bulk upsert with concurrency
+        with ThreadPoolExecutor(max_workers=concurrency) as executor:
+            futures = [executor.submit(self.upsert, document) for document in documents]
```
cbornet, line 894: I think `executor.map(self.upsert, documents)` can be used here.

astrapy/db.py
```diff
@@ -694,6 +697,51 @@ def insert_many(
...
+            list: The responses from the database after the batched insert operation.
+        """
+
+        # Split the documents into batches
+        batched_list = [
+            documents[i : i + batch_size] for i in range(0, len(documents), batch_size)
+        ]
+
+        # Function to insert a single batch
+        def insert_batch(batch: List[API_DOC], index: int) -> API_RESPONSE:
```
cbornet, line 726: Can't we use `self.insert_many` directly ? It's a lot of code just to add a debug log

astrapy/db.py
```diff
@@ -694,6 +697,51 @@ def insert_many(
...
+            partial_failures_allowed (bool, optional): Whether to allow partial failures in the batch.
+            batch_size (int, optional): Override the default insertion batch size.
+            concurrency (int, optional): The number of concurrent batch insertions.
+        Returns:
+            list: The responses from the database after the batched insert operation.
+        """
+
+        # Split the documents into batches
+        batched_list = [
+            documents[i : i + batch_size] for i in range(0, len(documents), batch_size)
```
cbornet, line 722: Perf improvement: this intermediate list is not needed. We could create the futures directly in the loop.

tests/astrapy/test_db_dml.py
```diff
@@ -324,6 +324,193 @@ def test_insert_many(writable_vector_collection: AstraDBCollection) -> None:
     assert isinstance(list(inserted_ids - {_id0, _id2})[0], str)
 
 
+@pytest.mark.describe("batched_concurrent_insert_many")
+def test_batched_concurrent_insert_many(
+    writable_vector_collection: AstraDBCollection,
+) -> None:
+    _id0 = str(uuid.uuid4())
+    _id2 = str(uuid.uuid4())
+    documents: List[API_DOC] = [
```
cbornet, line 333: This long list could be replaced by a repeated element.
Something like
```python
[{"_id": str(uuid.uuid4()), "name": "Abba"} for i in range(30)]
```

tests/astrapy/test_db_dml.py
```diff
@@ -324,6 +324,193 @@ def test_insert_many(writable_vector_collection: AstraDBCollection) -> None:
...
+        },
+        {
+            "_id": "90249293-07a7-43a0-a7f7-310a1ad91bad",
+            "name": "Ciccio",
+            "description": "The third in this list",
+            "$vector": [0.4, 0.3],
+        },
+    ]
+
+    response = writable_vector_collection.batched_concurrent_insert_many(documents)
```
cbornet, line 506: Set concurrency=2 ?

astrapy/db.py
```diff
@@ -694,6 +697,51 @@ def insert_many(
...
+        ]
+
+        # Function to insert a single batch
+        def insert_batch(batch: List[API_DOC], index: int) -> API_RESPONSE:
+            logger.debug(f"Processing batch #{index + 1} of size {len(batch)}")
+
+            return self.insert_many(
+                documents=batch,
+                options=options,
```
cbornet, line 731: If concurrency > 1, we could maybe set `options["ordered"] = False` as the insertions will be unordered anyway.

astrapy/db.py
erichare: I'll go with `chunked` based on the way we typically describe it, good suggestion, thank you!

astrapy/db.py
```diff
@@ -694,6 +697,51 @@ def insert_many(
...
+                partial_failures_allowed=partial_failures_allowed,
+            )
+
+        # Perform the bulk insert with concurrency
+        with ThreadPoolExecutor(max_workers=concurrency) as executor:
+            futures = [
+                executor.submit(insert_batch, batch, i)
+                for i, batch in enumerate(batched_list)
+            ]
+            response_list = [future.result() for future in futures]
```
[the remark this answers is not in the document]
hemidactylus: The Cassandra Python driver returns a list of `(Optional[result], Optional[error])` essentially acting like a `Either` (see [here](https://docs.datastax.com/en/developer/python-driver/3.29/api/cassandra/concurrent/#module-cassandra.concurrent)). I wonder if that would be a good model to adopt here.

astrapy/db.py
hemidactylus: Ideally one could collate the `insertedIds` and the `errors` field from each, both expected to be lists if present (I go by memory with field names).
I wonder however if that would backfire by hardcoding some assumptions on what the API returns that in the future could theoretically change.

astrapy/db.py
erichare: My only issue with that is, we chose insert_many because it's an API call. Here we're following the logic of basically wrapping the api call and producing chunks. So maybe chunked_upsert? to fit the convention of the other?

astrapy/db.py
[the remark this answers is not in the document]
cbornet: Why not. I find that it's not too hard to test if the result is an instance of `Exception`. But I don't have a strong opinion on this.

astrapy/db.py
cbornet: OK.

astrapy/db.py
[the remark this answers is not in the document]
hemidactylus: Same here - it can be left as simple as possible, no boolean is all good for me.

hemidactylus:
I'd like to raise a point which goes back to the suggestion to collate the response from the various API calls. I am not sure what the best design would be.

1. The [API specs](https://github.com/stargate/jsonapi/blob/main/docs/jsonapi-spec.md#insertmany-command-response) make it clear that the only fields are `status.insertedIds` (a list) and `errors` (a list), so collating seems to be as futureproof as it gets
2. In support of collating, the caller hardly needs to know the details of the splitting in chunks. They pass a list of 45 documents and would be surprised to get a list of 3 response JSON from the API, each about 20, 20, 5 insertions.
3. But then what about a single call failing with an exception? (not speaking of the case of "partial_failures_allowed", where the caller would not get an exception and know how to deal with the non-inserted entries).
 
To elaborate more on the last point: the caller sends 45 documents and the second API call raises an exception. What is more useful to the caller:
- getting back `[{...}, Exception, {...}]`, where they have to collate the insertedIds partial results and find out what went through (at the cost of being exposed to chunking, originally not a planned breach of abstraction)
- getting back a merged response (but then how would the Exceptions be passed back ?)
- seeing an exception being raised (but then how to find out which id got inserted, a very useful info) ?

Probably any choice other than keeping the response/exception separate by chunk is sacrificing some info, and I would avoid that  - at the cost of exposing the "surprising" chunked structure in the response item.

tests/astrapy/test_db_dml.py
hemidactylus: I have rewritten the whole test code.

PR #137, opened 2023-12-02
fix OPS_API_RESPONSE union type to include lists

hemidactylus:
I inadvertently gave a very silly definition of the OPS_API_RESPONSE, which did _not_ include `List[Any]` as one of the types (the DevOps API as a matter of fact does return top-level lists for some calls, such as `get_databases`).

Now I fixed it and to satisfy the type checker the whole of the ops methods are, correctly, moved to the OPS_API_RESPONSE return type.


erichare (approved):

erichare:
LGTM! :)

PR #143, opened 2023-12-11
Restore legacy REST API support

erichare:


erichare:
Per conversation with Preethi, closing this PR - all references to the old experience are going away. Note internally that users who rely on the rest module can `pip install astrapy==0.3.3`

PR #146, opened 2023-12-19
Add async API

cbornet:


tests/conftest.py
```diff
@@ -72,6 +73,14 @@ def db(astra_db_credentials_kwargs: Dict[str, Optional[str]]) -> AstraDB:
     return AstraDB(**astra_db_credentials_kwargs)
 
 
+@pytest.fixture
```
cbornet, line 76: We can't make it session-scoped because of https://github.com/pytest-dev/pytest-asyncio/issues/706#issuecomment-1838954312

erichare:
Note, tests failing due to secrets not being set in the origin repo. I tested this fairly extensively locally and its working FANTASTICALLY and this is such important functionality. Great job @cbornet

erichare (approved):

astrapy/db.py
```diff
@@ -979,3 +1708,247 @@ def truncate_collection(self, collection_name: str) -> AstraDBCollection:
...
+        namespace: Optional[str] = None,
+    ) -> None:
+        """
+        Initialize an Astra DB instance.
+        Args:
+            token (str, optional): Authentication token for Astra DB.
+            api_endpoint (str, optional): API endpoint URL.
+            namespace (str, optional): Namespace for the database.
+        """
+        self.client = httpx.AsyncClient()
```
cbornet, line 1729: We can't make it global as the AsyncClient pools connections and so must always be used from the same event loop.

PR #153, opened 2023-12-30
Use poetry to manage the project

cbornet:


pyproject.toml
```diff
@@ -0,0 +1,71 @@
+[tool.poetry]
+name = "astrapy"
+version = "0.6.3"
```
cbornet, line 3: At the moment, this is duplicated with `__version__` but there should be ways to only use this info.
For instance we can now get the installed package version with `importlib.metadata.version('astrapy')`

erichare: @cbornet This is great. could you look at the commit I just made? I think this will handle the issue. Let me know!

erichare (approved):

erichare:
LGTM! Thanks @cbornet

PR #154, opened 2024-01-01
Add optional pre-fetching of paginate results

### LINKED ISSUE #145: Pre-fetching of paginated results
Currently a new page is requested when the generator needs more rows.
It means that once a page of result has been processed, you have to wait for the new page of results to be fetched from Astra.
It would be nice to be able to specify a number of pages/rows that would be pre-fetched in background to increase the global throughput.

cbornet:
Fix #145


erichare (approved):
LGTM!

PR #162, opened 2024-01-10
a test collecting all non-equality operators in the filters to find

hemidactylus:
I am not even sure such a test belongs to astrapy, in a sense (astrapy's role as a passthrough for the payload is somewhat trivial here) - on the other hand it could help early discovery of regressions elsewhere. WDYT?

(this comes from an equivalent Java test, where the role of the sdk is more substantial since it's a fluent-interface "builder" kind of thing. I'm open to trash this or merge, whatever the consensus is).


erichare (approved):
LGTM, awesome!

cbornet:
If this is a test of the API more than astrapy itself, maybe this could go to another test file/suite ?
We want to align sync/async and it doesn't really make sense to duplicate this test in async.

PR #165, opened 2024-01-12
Align make_request and amake_request

cbornet:


hemidactylus (approved):
Thank you, I spotted that difference and then forgot :(

LGTM!

PR #169, opened 2024-01-13
Adjust tests to the ordered=False API default

hemidactylus:
The Data API beta7 has introduced the `ordered=False` default for insertMany.

The change in behaviour reverberate through several of the astrapy tests in not entirely trivial ways. This PR "fixes the tests" - which mostly amounts to

1. ensuring the correct `ordered` option is given explicitly when needed;
2. asserting on set equality and not list equality anymore in a few places, since for unordered inserts [the server is free to re-order the inserts and parallelize them for performance](https://github.com/stargate/jsonapi/commit/e25e8dab46015fb8d415b86ae21a57803d38ff27#diff-10a6655841047d3b5d397f09ba7216c68879c62a2efd3515e499d440796aee89).

### DIFF
--- a/tests/astrapy/test_db_dml.py
+++ b/tests/astrapy/test_db_dml.py
@@ -355,7 +355,8 @@ def test_chunked_insert_many(
         if isinstance(response, dict)
         for ins_id in response["status"]["insertedIds"]
     ]
-    assert inserted_ids0 == _ids0
+    # unordered inserts: this only has to be a set equality
+    assert set(inserted_ids0) == set(_ids0)
 
     response0a = writable_vector_collection.find_one(filter={"_id": _ids0[0]})
     assert response0a is not None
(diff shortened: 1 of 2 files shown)


tests/astrapy/test_db_dml.py
```diff
@@ -456,8 +462,15 @@ def test_concurrent_chunked_insert_many(
...
+        # the first doc must be pre-existing
+        # and the doc array size must be <= chunk size
+        # for this not to spoil the rest of the test
+        docs_for_error = documents0[0:1] + [{"_id": str(uuid.uuid4())}]
         _ = writable_vector_collection.chunked_insert_many(
-            documents1, chunk_size=3, concurrency=4
+            docs_for_error,
+            chunk_size=3,
+            concurrency=4,
+            options={"ordered": True},
```
cbornet, line 473: I wonder if it makes sense to allow to have both concurrency > 1 and ordered=True.
It might not give what the user expects...

cbornet (approved):

cbornet:
LGTM

tests/astrapy/test_db_dml.py
hemidactylus: Good point, I've been thinking about that a few times. But I would not go farther than issuing a warning if that condition is detected, wdyt? Being too prescriptive might get outdated faster possibly

PR #172, opened 2024-01-19
Feature/#166 slick truncate

### LINKED ISSUE #166: test and use the (newly-inserted) delete with empty payload for truncate
Test, expose and then look into using it in the integrations (e.g. langchain)

hemidactylus:
Closes #166 

This introduces usage of delete_many with empty payload for truncation instead of the drop-and-recreate previous flow. It also exposes a `truncate()` method at collection-level.

What is more, a massive overhaul of the tests is enabled by this, which optimizes DDL usage minimizing the amount of collection deletions through a stack of nested fixtures and truncations at the right layer in this stack.

Additionally, it is now possible to set `TEST_ASTRADBOPS=1` to skip the truncations entirely and have the tests complete in barely two minutes under current Astra DB conditions.

### DIFF
--- a/astrapy/db.py
+++ b/astrapy/db.py
@@ -848,18 +848,28 @@ class AstraDBCollection:
 
         return response
 
+    def clear(self) -> API_RESPONSE:
+        """
+        Truncate the collection, deleting all documents
+        Returns:
+            dict: The response from the database.
+        """
+        clear_response = self.delete_many(filter={})
+
+        if clear_response.get("status", {}).get("deletedCount") != -1:
+            raise ValueError(
+                f"Could not issue a truncation API command (response: {json.dumps(clear_response)})."
+            )
+
+        return clear_response
+
     def truncate(self) -> API_RESPONSE:
         """
         Truncate the collection, deleting all documents
         Returns:
             dict: The response from the database.
         """
-        truncate_response = self.delete_many(filter={})
-
-        if truncate_response.get("status", {}).get("deletedCount") != -1:
-            raise ValueError("Could not issue a truncate API command.")
-
-        return truncate_response
+        return self.clear()
 
     def delete_subdocument(self, id: str, subdoc: str) -> API_RESPONSE:
         """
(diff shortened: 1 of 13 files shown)


astrapy/db.py
```diff
@@ -848,6 +848,19 @@ def delete_many(self, filter: Dict[str, Any]) -> API_RESPONSE:
 
         return response
 
+    def truncate(self) -> API_RESPONSE:
```
cbornet, line 851: Maybe call it `clear` ? (same as cassio)

cbornet: We could also make `{}` the default filter of `delete_many` so `collection.delete_many()` would delete all.

hemidactylus: Agree on clear - I suggest however to keep truncate as an alias (it exists already and is theoretically used around)

As for the default in deleta_many, I am a bit afraid of the potential risk of wiping out inadvertently all your stuff, no?

cbornet: OK. Let's go for clear then.

tests/conftest.py
```diff
@@ -98,125 +98,198 @@ def invalid_db(
...
 ) -> AsyncIterable[AsyncAstraDBCollection]:
-    collection = await async_db.create_collection(
-        TEST_READONLY_VECTOR_COLLECTION,
-        dimension=2,
-    )
-
-    await collection.insert_many(VECTOR_DOCUMENTS)
+    """
+    This fixture piggybacks on its sync counterspart:
+    it must not do anything to the collection (test functions depend on both)
```
cbornet, line 215: Maybe `async_readonly_v_collection ` could have `readonly_v_collection`injection here instead of doing it in the test functions ?

hemidactylus:
Here is the final interface as discussed today:

- `AstraDB` does not implement any `clear_collection`, it still keeps `truncate_collection` with a clear deprecation notice urging to migrate (see next line)
- `AstraDBCollection` offers a `clear()` method which implements truncation. No use of the term "truncate" anywhere in the new API.

Likewise for ther async classes and methods.

Updated tests accordingly.

PR #177, opened 2024-01-25
Fix #176 - indexing support in collection creation

### LINKED ISSUE #176: support for "indexing" syntactic sugar in create collection
it's a brand new parameter at top-level in the request -> must be a named parameter in the method

erichare:
Needs a test... wasn't actually sure the best way to go about that especially in lieu of the recent issues with timeouts...

### DIFF
--- a/astrapy/db.py
+++ b/astrapy/db.py
@@ -1941,6 +1941,7 @@ class AstraDB:
         options: Optional[Dict[str, Any]] = None,
         dimension: Optional[int] = None,
         metric: Optional[str] = None,
+        indexing: Optional[Dict[str, Any]] = None,
     ) -> AstraDBCollection:
         """
         Create a new collection in the database.
@@ -1949,6 +1950,7 @@ class AstraDB:
             options (dict, optional): Options for the collection.
             dimension (int, optional): Dimension for vector search.
             metric (str, optional): Metric choice for vector search.
+            indexing (dict, optional): Indexing options for the collection.
         Returns:
             AstraDBCollection: The created collection object.
         """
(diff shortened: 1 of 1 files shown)


astrapy/db.py
```diff
@@ -1941,6 +1941,7 @@ def create_collection(
         options: Optional[Dict[str, Any]] = None,
         dimension: Optional[int] = None,
         metric: Optional[str] = None,
+        indexing: Optional[Dict[str, Any]] = None,
```
nicoloboschi, line 1944: shouldn't be better to expose the deny and allow parameters directly ? like

´´´
create_collection("test", allow_indexing=["field1"], deny_indexing=["field2"])
```

**Review by cbornet — APPROVED** [review]

**Comment by cbornet** [comment]
Is there a way to add a test ?

File: astrapy/db.py [hunk 2]
```diff
@@ -1987,7 +1989,11 @@ def create_collection(
         # Build the final json payload
         jsondata = {
             k: v
-            for k, v in {"name": collection_name, "options": options}.items()
+            for k, v in {
+                "name": collection_name,
+                "options": options,
+                "indexing": indexing,
```
hemidactylus, line 1995: Careful. "indexing" is not a new top-level param as it seemed initially.
It was later [confirmed](https://datastax.slack.com/archives/C04RPUB5JRM/p1706134743975189?thread_ts=1706101652.447869&cid=C04RPUB5JRM) that it is inside the "options" dict.

So...
1. not urgent to sugar the indexing into a friendlier way since one can take advantage of the feature right away
2. in case of separate indexing parameters to the method, similarly to 'dimension' and 'metric' it is necessary to check that it is not passed twice (once within the options and once as separate argument to create_collection)

And for what is worth, if sugar has to be, I would favor two parameters to the effect of allow and deny (and a check that no both are nonempty at the same time)

hemidactylus:
Everyone, **do not merge** as the mismatch above ("indexing" should be inside options) still stands.

Pro: not so urgent as options is supported already everywhere and users can take advantage of the indexing right away
Con: needs a small rework of the above (amounting to making it only sugar around the options stuff)

erichare:
Closing this for now. Will open a new PR for adding syntactic sugar to the indexing (allow, deny)

PR #178, opened 2024-01-30
Add tests for collection "indexing" settings

hemidactylus:
This PR adds two test on two new collections created with different non-default indexing settings: one is allow-list based, the other is deny-list based. Various filters are tested and it is verified they work (or don't) as expected.

I did not deem it necessary to double these tests in the async half of tests.

Tested in dev, 100% pass.
**Do not merge yet**, until the indexing makes it to prod.

Note: with these, the collections used in the test are now five. To avoid the "too many collections" error breaking the create-and-then-destroy-collection tests, these two are scoped to function (there are only 1 function using each of the two, anyway). In case deletions are suspended, no problem again since the tests at risk would be skipped.

### DIFF
--- a/tests/astrapy/test_db_dml.py
+++ b/tests/astrapy/test_db_dml.py
@@ -1323,3 +1323,53 @@ def test_insert_find_with_dates(
     assert response3 is not None
     document3 = response3["data"]["document"]
     assert document3 == expected_d_document
+
+
+@pytest.mark.describe("probe retrieval on a collection indexed with allowlist")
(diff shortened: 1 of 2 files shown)


hemidactylus:
Update: 1.0.0-RC3 is in prod now, which has full support for indexing. Tests indeed all pass.

hemidactylus:
Still having local test issues, out of caution I'll come back to this before merging, sorry

erichare (approved):
LGTM :)

PR #180, opened 2024-01-31
Fix #179 : alias for upsert calls

### LINKED ISSUE #179: alias upsert => upsert_one (with deprecation)
out of least-surprise, cf. insert_one and insert_many for instance.

erichare:
deprecate upsert(), use upsert_one() instead

### DIFF
--- a/astrapy/db.py
+++ b/astrapy/db.py
@@ -14,12 +14,11 @@
 from __future__ import annotations
 
 import asyncio
+import deprecation
 import httpx
 import logging
 import json
 import threading
-from warnings import warn
-
 
 from concurrent.futures import ThreadPoolExecutor
 from functools import partial
@@ -39,6 +38,7 @@ from typing import (
     AsyncGenerator,
 )
 
+from astrapy import __version__
 from astrapy.api import AsyncAPIRequestHandler, APIRequestHandler
 from astrapy.defaults import (
     DEFAULT_AUTH_HEADER,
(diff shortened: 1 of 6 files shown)


astrapy/db.py
```diff
@@ -1753,6 +1761,14 @@ async def delete_subdocument(self, id: str, subdoc: str) -> API_RESPONSE:
         return response
 
     async def upsert(self, document: API_DOC) -> str:
+        # Deprecated: call the upsert_one method
+        DEPRECATION_MESSAGE = "Method 'upsert' of AstraDBCollection is deprecated. Please switch to method 'upsert_one'."
+
+        warn(DEPRECATION_MESSAGE, DeprecationWarning, stacklevel=2)
```
cbornet, line 1767: The method `logging.warn` is deprecated. `logging.warning` should be used instead.

cbornet, line 1767: We could factorize in a `deprecation_warning` method ?

hemidactylus: This is `from warnings import warn` being used, probably good as is (no?)

hemidactylus: Agree- maybe a method decorator is even less intrusive. For that matter we might just use [this](https://pypi.org/project/deprecation/).

hemidactylus (approved):

astrapy/db.py
erichare: I like this idea! I'm going to push an update now which uses the deprecation module. it looks very dev friendly, and seems to work with our tests. We can easily update the messages etc and now they're just decorators on the function itself. See the update and let me know if you guys have any further feedback @cbornet @hemidactylus . Thanks :)

cbornet: Oh, sorry for the noise. It’s logging.warn which is deprecated…

PR #183, opened 2024-01-31
Add methods to_sync and to_async

cbornet:


astrapy/db.py
```diff
@@ -1872,6 +1882,15 @@ def __init__(
     def __repr__(self) -> str:
         return f'Astra DB[endpoint="{self.base_url}"]'
 
+    def to_async(self) -> AsyncAstraDB:
+        return AsyncAstraDB(
+            token=self.token,
+            api_endpoint=self.base_path,
```
hemidactylus, line 1888: I think this is not correct. The constructor receives `api_endpoint="https://<ID>-<REGION>.apps.astra.datastax.com"`, then builds `self.base_path="https://<ID>-<REGION>.apps.astra.datastax.com/v1/my_keyspace"`.
Using the latter as the `api_endpoint` for a subsequent constructor call would create an excess components in the new class' `self.base_path`.

In general, and to avoid having to follow the manipulations in the constructors (which tomorrow may change), it could be worth to capture all passed arguments to the constructors as they are and "freeze"[1] them as `self._constructor_args: Dict[str, Any]`, then these conversion methods could just blindly return `AstraDB(**self._constructor_args)`. There does not seem to be heavy or problematic stuff to freeze this way, no?

[1] mutable variables aside, which we may probably neglect in this scenario (these should be all strings, the problem should not be there).

cbornet: Oops. It’s base_url, not base_path

cbornet: fixed.

erichare:
LGTM!

PR #185, opened 2024-01-31
Fix #181 - use request handler for db ops

### LINKED ISSUE #181: AstraDB (+async) should use the request handler for correct management of 4xx and 5xx
Contrary to the collection classes, the AstraDB and AsyncAstraDB classes have this in their `request` method:

```
responsebody = cast(API_RESPONSE, response.json())
```

i.e. unprotected json coercion regardless e.g. of HTTP 500 errors and the like. Thus, the user would see the error as an unparseable JSON.

*Proposed solution*:
the `request` methods should use the same request handler as done in the collection classes.

(thanks @nicoloboschi for bringing this up!)

erichare:
Ensure we properly handle 400 and 500 HTTP errors by using the HTTP request handling classes previously developed, for AstraDB class

### DIFF
--- a/astrapy/db.py
+++ b/astrapy/db.py
@@ -50,9 +50,7 @@ from astrapy.defaults import (
 from astrapy.utils import (
     convert_vector_to_floats,
     make_payload,
-    make_request,
     http_methods,
-    amake_request,
     normalize_for_api,
     restore_from_api,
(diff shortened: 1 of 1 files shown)


nicoloboschi (approved):
LGTM

PR #187, opened 2024-01-31
Fix #186 - token and api endpoint not optional

### LINKED ISSUE #186: Revise Optional[str] constructor parameters when it does not make sense
Why `token: Optional[str]` and `api_endpoint` likewise when then it's a ValueError not passing it? Probably a leftover of old times, can be made more adherent to expectations to help developers.

erichare:
Note that this applies specifically to the AstraDB (and async version) classes - for the Collection classes, one can provide either an astra_db instance already initialized OR pass a (token + endpoint). So those have been left as is.

### DIFF
--- a/astrapy/db.py
+++ b/astrapy/db.py
@@ -1867,8 +1867,8 @@ class AstraDB:
         """
         Initialize an Astra DB instance.
         Args:
-            token (str, optional): Authentication token for Astra DB.
-            api_endpoint (str, optional): API endpoint URL.
+            token (str): Authentication token for Astra DB.
+            api_endpoint (str): API endpoint URL.
             namespace (str, optional): Namespace for the database.
         """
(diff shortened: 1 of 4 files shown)


hemidactylus (approved):

erichare:
> LGTM, just a question: is there any specific reason for replacing the `**{...}` unpacking of the credentials in the tests with explicit handling -- other than readability/aesthetics? (I mean, were there typing issues or something?)

Yep, the reason was mypy. It didn’t like the fact that we are unpacking the credentials, because then it’s thinking that the required parameters are actually optional. I tried some kind of hacky workarounds, but in the end thought this was okay. If the universe of optional parameters expands though, we can restore the unpacking syntax for those, and just keep the required ones as explicitly provided.

hopefully that makes sense

PR #192, opened 2024-02-05
Adjust and test for TRACE logging

### LINKED ISSUE #191: check TRACE logging
since this is what I see when enabling level 5 (the custom-defined "trace"):

trace_logging

hemidactylus:
Fixes #191 .

Besides fixing the trace-logging command (which was erroring due to there being an `*args` where just `args` should have been), this also adds a specific test to see that such logging is working and indeed obeyed.

All test _would_ pass - at least locally for me, except the (unrelated) `__version__` autodetect logic fails on my poetry environment (detecting a much-too-old 0.6.2), so that the expected deprecation warnings are ("not yet") raised and a few tests fail because of that.


erichare (approved):
LGTM thank you so much Stefano! Bad typo on my part

PR #198, opened 2024-02-07
Full API to manage the caller identity

### LINKED ISSUE #193: Enable an API for the caller to set caller_name/caller_version
For the AstraDB, AstraDBCollection and their async counterparts.

This means (a) copy() methods, and (b) set_caller methods.

To ensure all invocations are called, and less maintenance, these will be class settings. It is up to the caller to use copy() if if can detect the classes are not used exclusively by it.

hemidactylus:
Closes #193.

this PR builds on the internal refactoring and exposes a consistent interface for code using these classes (AstraDB, AstraDBCollection, AsyncAstraDB, AsyncAstraDBCollection, AstraDBOps) to set the "caller_name" and "caller_version" in a way that is then reflected in the User-Agent header for all subsequent requests originating from the instance.

The API offers two options:
- (optional) constructor parameters
- a separate setter method

The PR adds testing for the to_sync and to_async methods (which was lacking and is somewhat related to this PR as a sibling of the newly-introduced copy() methods), as well as unit testing for all the caller-identity functionalities and http-testing for the user-agent header induced by the caller identity.

A copy() method was indeed introduced to allow for the pattern by the caller whereby they make a copy and then feel entitled to set_caller() on them without interfering with other usages of the passed client.

Also a "reasonable" __eq__ method is introduced for all these classes, mainly to make tests around this feature easier to write.

As it turns out, subtleties abound. Here is a summary of the design choices taken:
- if creating collections passing a client, the latter's caller is default for the former;
- when a collection is spawned from a client, inheritance of caller takes place;
- when set_caller is invoked on a collection, it is further propagated upward to the enclosed client;

### DIFF
--- a/astrapy/db.py
+++ b/astrapy/db.py
@@ -88,26 +88,45 @@ class AstraDBCollection:
             api_endpoint (str, optional): API endpoint URL.
             namespace (str, optional): Namespace for the database.
             caller_name (str, optional): identity of the caller ("my_framework")
+                If passing a client, its caller is used as fallback
             caller_version (str, optional): version of the caller code ("1.0.3")
+                If passing a client, its caller is used as fallback
         """
-        self.caller_name = caller_name
-        self.caller_version = caller_version
         # Check for presence of the Astra DB object
         if astra_db is None:
             if token is None or api_endpoint is None:
                 raise AssertionError("Must provide token and api_endpoint")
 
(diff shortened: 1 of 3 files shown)


erichare (approved):
This is fantastic. LGTM @hemidactylus :) going to be a big improvement

PR #200, opened 2024-02-08
Fix #175 : if concurrency and ordered, warn.

### LINKED ISSUE #175: Warn against concurrency in batching and ordered=true
This combination does not make much sense, and while not forbidding it entirely astrapy should gently issue a warning when detecting it

erichare:


hemidactylus (approved):
LGTM

PR #204, opened 2024-02-09
Fix #203 : Implementation for update_many

### LINKED ISSUE #203: Implement `update_many`
The data API offers the methods `update_many` that allows to update a subset of properties for different record matching a criteria

Documentation
https://docs.datastax.com/en/astra/astra-db-vector/api-reference/data-api-commands.html#update-many-matching-documents-with-set

Sample:
```json
{ 
  "updateMany": {
    "filter": {"status" : "active" },
    "update" : {"$set" : { "status" : "inactive"}}
  }
} 
```

As of today this feature is not exposed in the astrapy.

erichare:


tests/astrapy/test_db_dml.py
```diff
@@ -790,6 +790,26 @@ def test_update_one_create_subdocument_novector(
     assert response["data"]["document"]["name"] == "Eric"
 
 
+@pytest.mark.describe("update_many to create a subdocument, not through vector")
+def test_update_many_create_subdocument_novector(
+    writable_v_collection: AstraDBCollection,
```
hemidactylus, line 795: Replace with `empty_v_collection` fixture otherwise you may have surprises with the count at the end (writable gives you no guarantee there _won't_ be other records shadowing your failing test at the last assert).

erichare: Thanks @hemidactylus . great point. Updated!

PR #207, opened 2024-02-13
Fix #206 upgrade httpx and related deps

### LINKED ISSUE #206: Update to "httpx" = "0.26.0"

erichare:


erichare:
@hemidactylus i know we discussed whether there was some issue with httpx 0.26.0. Sort of testing the waters with this. This PR actually does upgrades on dependency packages across the board, nothing major so far as i can tell but we'll want to certainly test more before merging. Just wanted to kind of get the ball rolling on it while we're in this rapid dev phase.

PR #216, opened 2024-02-23
From @synedra for unsupported methods

erichare:
This adds a list of unsupported Mongo methods for each Astra class


synedra (approved):
These look fine and are in sync with Cedrick's document

PR #217, opened 2024-02-23
Database metadata through name, id, region

synedra:
Added dbops calls to the astraDB creation to gather name, id, and region.  Exposed through attributes:
astradb.name
astradb.region
astradb.id

Additionally added "name, full_name" to collections as well as the database
col.name
col.fullname
col.database

Also added client_options at the DB level.  It is set outside the class so we could expose it at the collection level or within the asyncs.

Also added collection.drop() and changed db.delete_collection() to db.drop_collection()

Tests for everything, everything is passing.

Document with needed items: https://docs.google.com/spreadsheets/d/1ykhgGAqgxdnGJfOWevGDCcD3K6cZPxlGtN99bZTt_S4/edit#gid=2080544998

### DIFF
--- a/astrapy/db.py
+++ b/astrapy/db.py
@@ -952,6 +952,18 @@ class AstraDBCollection:
             )
 
         return clear_response
+    
+    def drop(self) -> API_RESPONSE:
+        """
+        Drop the collection, deleting all documents
+        Returns:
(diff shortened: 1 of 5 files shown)


hemidactylus:
This is not meant for `master`, should be closed.

PR #218, opened 2024-02-26
Adding read preference to the collection

synedra:
Adding read preference as an attribute on the collection.


hemidactylus:
Isn't this PR to be closed because it is subsumed into #219 entirely?

PR #219, opened 2024-02-26
Pymongo codec options

synedra:
Adding codec_options to the collection level.  Currently an empty list because we don't support those items.

### DIFF
--- a/astrapy/db.py
+++ b/astrapy/db.py
@@ -113,6 +113,9 @@ class AstraDBCollection:
         self.collection_name = collection_name
         self.base_path = f"{self.astra_db.base_path}/{self.collection_name}"
 
+        # Set read concern to "NEAREST"
+        self.read_preference = "NEAREST"
+
     def __repr__(self) -> str:
         return f'AstraDBCollection[astra_db="{self.astra_db}", collection_name="{self.collection_name}"]'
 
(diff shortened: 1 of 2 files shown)


erichare:
@synedra i think this might include the read preference code from your other PR. I left a question there for you... im assuming this is all pymongo related now... if so i'd just say either we close the other PR and only use this one, or maybe change this one to not have the read preference change from the other one?

synedra:
Let’s let it go (reject it), as we’re moving things onto the m1 branch instead.

Thanks,
K

astrapy/db.py
```diff
@@ -113,6 +113,10 @@ def __init__(
         self.collection_name = collection_name
         self.base_path = f"{self.astra_db.base_path}/{self.collection_name}"
 
+        # Set read concern to "NEAREST"
```
hemidactylus, line 116: Is read concern the same as read preference? The mongo [docs](https://pymongo.readthedocs.io/en/stable/api/pymongo/collection.html#pymongo.collection.Collection.read_preference) seem to imply that they are not.
In that case the comment may seem confusing here

hemidactylus:
I think this PR is superseded by (what will be the one from branch) `mongodb-database-attributes`, correct?

hemidactylus:
> I think this PR is superseded by (what will be the one from branch) mongodb-database-attributes, correct?

I mean, #231 is making this one obsolete, correct @synedra  ?

synedra:
Closed this one and the other, opened a new one to the right branch.

K

PR #221, opened 2024-02-28
enhanced copy methods for astrapy objects

hemidactylus:
This PR adds the possibility to supply override to any parameter during calls to the `copy()` method (for AstraDB, AstraDBCollection, +the two async counterparts, and AstraDBOps).

Building on this, care is taken that the constructor for AstraDBCollection (+async) does not silently discard some of the parameters when the `astra_db` is passed (namely namespace, token and endpoint), rather treats them as overrides on the passed astra_db.

Full testing is also added.

### DIFF
--- a/astrapy/db.py
+++ b/astrapy/db.py
@@ -104,6 +104,13 @@ class AstraDBCollection:
                 caller_name=caller_name,
                 caller_version=caller_version,
             )
+        else:
+            # if astra_db passed, copy and apply possible overrides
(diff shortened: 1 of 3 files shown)


erichare:
Fantastic @hemidactylus . i've been out as you know and seeing this made me very happy :D and as always a very thorough test. LGTM and thank you for taking care of this!

erichare (approved):

PR #225, opened 2024-02-28
Mongodb collection drop function

synedra:
Adding a function for a collection to drop itself: collection.drop(COLLECTION_NAME).

Added a test to validate that it's working.  It *seems* like a ddl sort of thing so that's where I put it.

### DIFF
--- a/astrapy/idiomatic/collection.py
+++ b/astrapy/idiomatic/collection.py
@@ -90,6 +90,9 @@ class Collection:
             caller_version=caller_version,
         )
 
+    def drop(self, collection_name: str) -> API_RESPONSE:
+        return self.astra_db.delete_collection(collection_name)
+
     @unsupported
     def find_raw_batches(*pargs: Any, **kwargs: Any) -> Any: ...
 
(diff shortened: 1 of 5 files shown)


astrapy/idiomatic/collection.py
```diff
@@ -216,6 +216,9 @@ def delete_many(
                 f"(gotten '${json.dumps(dm_response)}')"
             )
 
+    def drop(self, name_or_collection: Union[str, AsyncCollection]) -> None:
```
hemidactylus, line 219: As a method on "collection", it has no "name_or_collection" argument (see [ref](https://pymongo.readthedocs.io/en/stable/api/pymongo/collection.html#pymongo.collection.Collection.drop)), it being implied that the one to drop is self.

note: The Mongo docs use the term "alias" pretty liberally here IMO :D

astrapy/idiomatic/collection.py
```diff
@@ -445,6 +448,10 @@ async def delete_many(
                 f"(gotten '${json.dumps(dm_response)}')"
             )
 
+    async def drop(self, name_or_collection: Union[str, AsyncCollection]) -> None:
+        return await self._astra_db_collection.astra_db.delete_collection(name_or_collection)
```
hemidactylus, line 452: same for the async, and incidentally I am surprised if this typechecks ...

hemidactylus:
Can you confirm this is subsumed by #231 (as it seems to me?) In that case, please close.

PR #228, opened 2024-02-29
feat: add ragstack-ai to user-agent if present

nicoloboschi:


hemidactylus:
Tests fail because the PR comes from another fork.
I did them locally, all green.

PR #230, opened 2024-02-29
Mongodb database attributes

synedra:
Adding attributes to the database per Clun's spreadsheet

- name
- dbid
- database (an object with name, id and region
- region
- client_options
- codec_options (empty list as per Cedrick)
- read_preference (Cedrick said this will always be NEAREST


hemidactylus (commented):
This should most likely be a PR to `pymongo-convergence-m1` rather than `master`.

PR #231, opened 2024-02-29
Mongodb database attributes

synedra:
Adding attributes to the database per Clun's spreadsheet

name
dbid
database (an object with name, id and region
region

Adding a collection.drop() function to match mongo

Going to the right branch this time.

### DIFF
--- a/astrapy/idiomatic/collection.py
+++ b/astrapy/idiomatic/collection.py
@@ -90,6 +90,9 @@ class Collection:
             caller_version=caller_version,
         )
 
+    def drop(self, collection_name: str) -> API_RESPONSE:
+        return self.astra_db.delete_collection(collection_name)
+
     @unsupported
     def find_raw_batches(*pargs: Any, **kwargs: Any) -> Any: ...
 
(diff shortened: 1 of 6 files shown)


astrapy/idiomatic/collection.py
```diff
@@ -216,6 +216,9 @@ def delete_many(
                 f"(gotten '${json.dumps(dm_response)}')"
             )
 
+    def drop(self, name_or_collection: Union[str, AsyncCollection]) -> None:
```
hemidactylus, line 219: As a method on collection this has no arguments, it is implied that [self is going to be dropped](https://pymongo.readthedocs.io/en/stable/api/pymongo/collection.html#pymongo.collection.Collection.drop).

hemidactylus:
If you pull the latest and adjust the conflicts, I think you'll be able to do all async tests. I fixed a few minor problems with the async fixtures in my last push.

astrapy/idiomatic/database.py
```diff
@@ -91,6 +92,35 @@ def __init__(
...
+            
+            details = astraDBOps.get_database(database=self.dbid)
+            self.name = details['info']['name']
+            self.region = details['info']['region']
+            self.database = {'id': self.dbid, 'name': self.name, 'region': self.region}
+
+        else:
+            self.name = ""
+            self.region = ""
+            self.database = ""
```
hemidactylus, line 122: Perhaps a better typing would be less surprising, i.e. `self.database: Optional[Dict[str, Any]]` or something like that.

PR #263, opened 2024-03-26
Do not accept NaNs in serializing for the API

hemidactylus:
`Nan` is not a valid JSON token. Sadly for some reason the JSON serialization in python [does not complain](https://stackoverflow.com/questions/6601812/sending-nan-in-json?rq=3), but astrapy should since the Data API is strictly JSON.

This PR addresses this problem.


hemidactylus:
closing, bad target branch!

PR #287, opened 2024-06-26
Fixed bug in parsing endpoint domain names containing hyphens

bradfordcp:
I recently encountered an issue with a user who is running the Data API against a DSE cluster. This Data API is behind a domain name containing a dash. For example:

```python
from astrapy import DataAPIClient
from astrapy.constants import Environment

client = DataAPIClient(token="Cassandra:...:...", environment=Environment.DSE)
db = client.get_database_by_api_endpoint("https://example-domain.com", namespace="default_namespace")
```

would fail as astrapy is unable to parse the endpoint. This PR adjusts the regular expression to include hyphens. This could technically result in invalid domain names being parsed (hyphens are not allowed at the start of domain names), but solves the immediate issue. I've also added tests to validate the endpoint parsing is working as expected.

### DIFF
--- a/astrapy/admin.py
+++ b/astrapy/admin.py
@@ -75,7 +75,7 @@ api_endpoint_parser = re.compile(
     r".datastax.com"
 )
 
-generic_api_url_matcher = re.compile(r"^https?:\/\/[a-zA-Z0-9.]+(\:[0-9]{1,6}){0,1}$")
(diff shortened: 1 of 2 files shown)


hemidactylus:
Thank you for spotting that and offering a ready-made solution!
(no worries about CI failing, it is a secrets problem - running locally on my machine got all green)

PR #299, opened 2024-08-20
Fix find prefetching

cbornet:


astrapy/core/db.py
```diff
@@ -54,6 +55,74 @@
...
+        self.request_method = request_method
+        self.options = options
+        self.raw_response_callback = raw_response_callback
+        self.initialised = threading.Event()
+        self.stop = threading.Event()
+        self.thread = threading.Thread(
+            target=_PrefetchIterator.queued_paginate, args=(weakref.proxy(self),)
+        )
+        self.thread.start()
+        while not self.initialised.is_set():
```
hemidactylus, line 76: wouldn't this loop consume many resources without a (tiny) sleep in its body or something?
I think `Event.is_set()` does not involve any thread-synch, it would continuously read and read and read (?)

astrapy/core/db.py
```diff
@@ -54,6 +55,74 @@
...
+    def __iter__(self) -> Iterator[API_DOC]:
+        return self
+
+    @staticmethod
+    def queue_put(
+        q: queue.Queue[Optional[API_DOC]],
+        item: Optional[API_DOC],
+        stop: threading.Event,
+    ) -> None:
+        while not stop.is_set():
```
hemidactylus, line 91: same as above, doesn't `is_set` immediately return? (if so, this might be a resource-heavy loop)

astrapy/core/db.py
cbornet: I copied that from somewhere.
But `self.initialised.wait()` would probably be much better.

cbornet: done.

tests/core/test_db_dml_pagination.py
```diff
@@ -30,6 +31,7 @@
 
 FIND_LIMIT = 183  # Keep this > 20 and <= N to actually put pagination to test
 PREFETCHED = 42  # Keep this > 20 and <= FIND_LIMIT to actually trigger prefetching
+ROWS_PER_PAGE = 20
```
hemidactylus, line 34: This does not seem to be used anywhere.

tests/core/test_db_dml_pagination.py
```diff
@@ -43,16 +45,46 @@
...
+        assert caplog.text.count("HTTP Request: POST") == 0
+
+    paginated_documents = list(paginated_documents_it)
     paginated_ids = [doc["_id"] for doc in paginated_documents]
     assert all(["$vector" not in doc for doc in paginated_documents])
     assert len(paginated_ids) == FIND_LIMIT
     assert len(paginated_ids) == len(set(paginated_ids))
+
+
+def test_prefetched_thread_terminated(
```
hemidactylus, line 74: would it be ok to add an async one in the async twin test module? (perhaps does not add much, but...)

```
import asyncio
```

and later 

```
async def test_prefetched_thread_terminated(
    async_pagination_v_collection: AstraDBCollection,
) -> None:
    options = {"limit": FIND_LIMIT}
    projection = {"$vector": 0}

    paginated_documents_agen = async_pagination_v_collection.paginated_find(
        projection=projection, options=options, prefetched=PREFETCHED
    )

    assert await anext(paginated_documents_agen) is not None
    del paginated_documents_agen
```

tests/core/test_db_dml_pagination.py
cbornet: good catch

tests/core/test_db_dml_pagination.py
cbornet: The async counterpart doesn't spawn a thread.
It creates an async task which arguably has the same default of leaking.
This can be considered in another PR ?

tests/core/test_db_dml_pagination.py
cbornet: done

tests/core/test_db_dml_pagination.py
hemidactylus: Of course!

hemidactylus (approved):
I highly admire your solution. LGTM

PR #300, opened 2024-08-21
Restore docker-compose usage + upgrade HCD images

hemidactylus:
without committing to python 3.9, as a modern testcontainers would require.


cbornet:
An alternative would be to drop support for Python 3.8.
It will be EOL in October anyway: https://devguide.python.org/versions/

PR #309, opened 2024-08-28
HTTP Status Errors: enrich-and-reraise

hemidactylus:
This PR introduces a new astrapy exception in the class hierarchy, `DataAPIHttpException`, sibling to `DataApiResponseException`.

The new one represents 4xx/5xx responses from the API. Since in some cases these responses still contain useful info in the standard `{"errors": [...]}` format, which would be omitted in string representation of the httpx exception, now astrapy does the following:
- catches 4xx and 5xx exceptions as httpx errors
- repackages them as DataAPI exceptions, enriching them with the "error descriptors" as it does for "response exceptions" already;
- raises them.

Note that the new error class (see `pictures/astrapy_exceptions.png`) inherits from both `DataAPIException` and the httpx class: this, while not presenting relevant "diamond-type" multiple-inheritance problems, allows current flows based on catching the httpx exceptions to continue working.


astrapy/exceptions.py
```diff
@@ -211,6 +211,77 @@ class DataAPIException(ValueError):
...
+
+    text: Optional[str]
+    error_descriptors: Optional[List[DataAPIErrorDescriptor]]
+
+    def __init__(
+        self,
+        text: Optional[str],
+        *,
+        httpx_error: httpx.HTTPStatusError,
+        error_descriptors: List[DataAPIErrorDescriptor],
```
erichare, line 238: Should this be Optional here?

erichare (approved):

astrapy/exceptions.py
toptobes: Wouldn't `[]` convey the same information as `Nothing`? (***in this context***)

hemidactylus: Agree; and agree.
The Optional was an oversight from me, sorry and thank you for spotting this!
In line with the counterpart exception (DataAPIResponseException), there's no Optional and it is `[]` if nothing.

thanks!

PR #314, opened 2024-09-16
Complete removal of "core" from the idiomatic API

hemidactylus:
This PR removes the last big item of technical debt from astrapy. The whole client API is rewritten not to use any of the "core" (or pre-1.0) library, relying instead of the `APICommander` for all of its interactions with both the Data and DevOps APIs.

Additionally, with version 1.5 in mind, a deprecation notice is put in place for the whole "core" submodule (slated for removal in v2).

The rewriting of the whole client logic to do without "core" has implications in several places (see the CHANGES file for a quick map); but, notably, it allows to entirely remove the catch-and-reraise mechanism that used to be in place for most request-issuing methods in order to recast exceptions from core into "idiomatic" exceptions. This, incidentally, allows for a better control of the return types from said methods.


erichare (approved):
Exciting!! LGTM

PR #343, opened 2025-03-10
fixed spelling error in example

epinzur:


hemidactylus:
Thanks for spotting that, Eric!
I have fixed it manually in a giant branch that will become astrapy 2.0.0 soon, if you don't mind.
But still 100% of the credit to you for finding this one!

PR #349, opened 2025-05-15
Switch to "uv"

### LINKED ISSUE #325: Migrate project to leverage Astral UV
See [here](https://astral.sh/blog/uv-unified-python-packaging) for the benefits (over current things like `pip`, `poetry`, etc.,)! We could even continue to leverage the same `pyproject.toml` file to manage the project dependencies, etc.

Othere References:
- [some deep-dive blog](https://www.saaspegasus.com/guides/uv-deep-dive/)

hemidactylus:
Closes #325


pyproject.toml
```diff
@@ -1,70 +1,69 @@
...
-pytest-testdox = "~3.1.0"
-pytest = "~8.0.0"
-python-dotenv = "~1.0.1"
-pytest-httpserver = "~1.0.8"
-testcontainers = "~3.7.1"
-ruff = "^0.6.8"
-types-toml = "^0.10.8.7"
-blockbuster = "~1.5.5"
+[project.optional-dependencies]
+dev = [
```
cbornet, line 42: It would be better to put dev dependencies in the dev group.
See https://docs.astral.sh/uv/concepts/projects/dependencies/#development-dependencies
Then when you install, you do it with `uv sync --dev`

pyproject.toml
```diff
@@ -1,70 +1,69 @@
...
+    "setuptools >= 61.0",
+    "testcontainers ~= 3.7.1",
+    "types-toml >= 0.10.8.7,<1.0.0",
+    "twine >= 4.0.0"
+]
 
 [build-system]
-requires = ["poetry-core>=1.0.0"]
-build-backend = "poetry.core.masonry.api"
+requires = ["setuptools >= 61.0"]
```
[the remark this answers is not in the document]
cbornet: Here is the conf for BlockBuster using Hatchling if that helps: https://github.com/cbornet/blockbuster/blob/main/pyproject.toml
I think you would just need
```toml
[tool.hatch.build.targets.wheel]
packages = ["astrapy"]

[build-system]
requires = ["hatchling"]
build-backend = "hatchling.build"
```

Makefile
```diff
@@ -27,33 +31,31 @@ format-fix-tests: FMT_FLAGS=
...
+	uv run pytest tests/base/unit -vv
 
 docker-test-integration:
-	DOCKER_COMPOSE_LOCAL_DATA_API="yes" poetry run pytest tests/base -vv
+	DOCKER_COMPOSE_LOCAL_DATA_API="yes" uv run pytest tests/base -vv
 
 build:
 	rm -f dist/astrapy*
-	poetry build
+	uv run python -m build
```
cbornet, line 44: I think it's `uv build`

pyproject.toml
```diff
@@ -1,70 +1,69 @@
...
-ruff = "^0.6.8"
-types-toml = "^0.10.8.7"
-blockbuster = "~1.5.5"
+[project.optional-dependencies]
+dev = [
+    "blockbuster ~= 1.5.5",
+    "build >= 1.0.0",
+    "faker ~= 23.1.0",
+    "mypy ~= 1.9.0",
+    "ruff >= 0.6.8,<1.0.0",
```
cbornet, line 47: nit: ruff makes breaking changes between minor versions. Hence why there are formatting changes brought by this PR.
I would suggest using `ruff >= 0.11.9,<0.12`

pyproject.toml
```diff
@@ -1,70 +1,69 @@
...
+    "pytest ~= 8.0.0",
+    "pytest-asyncio ~= 0.23.5",
+    "pytest-cov ~= 4.1.0",
+    "pytest-testdox ~= 3.1.0",
+    "python-dotenv ~= 1.0.1",
+    "pytest-httpserver ~= 1.0.8",
+    "setuptools >= 61.0",
+    "testcontainers ~= 3.7.1",
+    "types-toml >= 0.10.8.7,<1.0.0",
+    "twine >= 4.0.0"
```
cbornet, line 58: I think twine is not needed.
You should be able to publish with `uv publish`

PR #399, opened 2026-02-23
feat: More robust release workflow for PyPi

### LINKED ISSUE #382: Improvements in the release workflow (and workflows in general)
The following should be done as a second sweep:
 
- stop the flow early if version foun on pypi already (to avoid running long testing for nothing)
- Adopt the "setup uv" action everywhere in the other flows
- figure out about 'attestations' and re-enable if ok to do so (two places: in release and _test_release)
- Completely revoke, once full functionality verified with an actual release, any PyPI release key for astrapy

erichare:
Fixes #382 

# More robust release workflow for PyPI

## Summary

Second-sweep improvements to the publish-and-release GitHub Actions workflow, plus standardization of CI tooling across all workflow files.

## Changes

🚀 Release workflow — early exit for existing versions
•  Added a check-pypi step in the build job that queries the PyPI JSON API (https://pypi.org/pypi/{pkg}/{version}/json) to detect if the version is already published
•  All downstream jobs (test-pypi-publish, pre-release-checks, pre-release-unit-lowest-python, publish, mark-release) are now gated with if: needs.build.outputs.version-exists == 'false', avoiding expensive test runs when there's nothing to release

🔧 Standardized uv_setup composite action across CI
•  Migrated lint.yml, local.yml, main.yml, and unit.yml from the legacy setup-python@v2 + pipx install uv + make venv pattern to the shared .github/actions/uv_setup composite action + uv sync --dev
•  Removed the now-unnecessary LEGACY_INSERTMANY_BEHAVIOUR_PRE2193 secret requirement from main.yml and codecov_aggregator.yml
•  Uncommented HEADER_RERANKING_API_KEY_NVIDIA in local.yml

🔒 Attestations re-enabled
•  Switched attestations: false → attestations: true in both release.yml (production PyPI) and _test_release.yml (test PyPI)
•  Added a NOTE comment as a reminder to revoke legacy PyPI API tokens once this trusted-publishing workflow is verified with an actual release

🧹 Minor cleanup (review-driven)
•  Removed leftover "TEST FOR NOW" from the publish step name
•  Added a 30s timeout to the PyPI version-check HTTP request to prevent workflow hangs
•  Removed unused import json from the check-pypi inline script


erichare:
Superseded by #411 — rebased onto the latest `main` and moved to a branch in `datastax/astrapy` (instead of my fork). Closing in favor of that PR.

PR #415, opened 2026-06-09
Upgrade CI/CD (replace OpenAI->Voyage; suppress OpenAI from extended vec.ze)

sl-at-ibm:
Pending a valid OpenAI API Key to use, this PR:

- factors in a single place the choice of the (auth-requiring) vectorize provider used in the workflow CI/CD process;
- replaces its setting to OpenAI with VoyageAI (for which we have a working API Key)
- while we're at it, suppresses (temporarily) OpenAI in its entirety from the "vectorize" portion of the integration tests (which are not run by the github automation, but still)


sl-at-ibm:
@erichare I'm asking for your review mostly to let you know that this will unblock the CI/CD process and usher in a new flood of improvements in the form of subsequent PRs. So, all in all, very good news (and a much-needed refactor of the integration tests around vectorize).

erichare (approved):
Awesome, LGTM @sl-at-ibm .  Very good news indeed!

sl-at-ibm:
Well, it turned out this IT upgrade had a lot of hidden work that became apparent only after starting the overhaul, but I should be nearing completion now!
Unfortunately still some flakiness on the Astra integration test side (timeouts, metadata sync issues and such), but I'll work out a way.

sl-at-ibm:
Force-merging because all of IT has succeeded except the downloading of coverage json from latest `main` commit (to compute codecov diff). "Artifact has expired".

Should be healed once this gets to main, CI runs again, and stores the new main-coverage-json for subsequent PRs.

PR #421, opened 2026-06-16
Enrich create_database method expressiveness

### LINKED ISSUE #408: Database Creations enhancements in Astra
### **Description**  

Currently, when creating a database in Astra via the client, the available options are very limited. Specifically, database creation is constrained to **vector databases only**, with no flexibility to choose other database types.

However, recent Astra platform evolutions introduce new capabilities that are not yet exposed at the client level, creating a gap between platform features and client usability.

### **Current Behavior**  
- Database creation is restricted to vector databases  
- Only basic parameters such as **name** and **location** are supported  
- No support for newer configuration options  

### **Expected Behavior**  
The client should support more flexible database creation aligned with Astra’s latest capabilities, including:

- Ability to create databases with different **flavors/types**, not limited to vector databases  
- Support for future availability of the **Data API across all database types**  
- Ability to assign a database to a **PCU group at creation time** using a new parameter (e.g., `pcuGroupUUID`)

### **Proposed Solution**  
Refactor the database creation API to introduce a richer request model:

- Introduce a `DatabaseCreationRequest` object encapsulating:
  - Database name  
  - Location  
  - Database type/flavor  
  - PCU group assignment (`pcuGroupUUID`)  
  - Additional future configuration options and feature flags  

This approach improves extensibility and ensures alignment with evolving Astra platform capabilities.

### **Impact**  
- Limits adoption of new Astra features  
- Prevents usage of non-vector database types  
- Reduces flexibility for users managing infrastructure through the client

sl-at-ibm:
This PR expands the capabilities of the AstraDBAdmin's `create_database` (+ `async_`) by exposing the payload settings so far hardcoded in the clients (db type + capacity units + tier), and the newly-introduced "pcu group uuid".

This is done not with additional keyword arguments, one per setting, rather with the introduction of a `DatabaseDefinition` object with all properties (minus `name`, kept separate as customary in client createXYZ method signatures) and a new `definition` parameter to `create_database`, mutually exclusive with the (pre-existing) corresponding keyword arguments.

Tests and examples in the docstring will explain the rest.

Fixes #408 .

### DIFF
--- a/astrapy/admin/admin.py
+++ b/astrapy/admin/admin.py
@@ -48,6 +48,7 @@ from astrapy.info import (
     AstraDBAdminDatabaseInfo,
     AstraDBAvailableRegionInfo,
     AstraDBDatabaseInfo,
+    DatabaseDefinition,
     FindEmbeddingProvidersResult,
     FindRerankingProvidersResult,
 )
@@ -1065,8 +1066,9 @@ class AstraDBAdmin:
         self,
         name: str,
         *,
-        cloud_provider: str,
-        region: str,
+        definition: DatabaseDefinition | None = None,
+        cloud_provider: str | None = None,
+        region: str | None = None,
         keyspace: str | None = None,
         wait_until_active: bool = True,
         database_admin_timeout_ms: int | None = None,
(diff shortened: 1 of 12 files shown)


astrapy/admin/admin.py
```diff
@@ -1298,17 +1351,59 @@ async def async_create_database(
...
+            >>> # call pattern using the 'definition' parameter:
+            >>> my_custom_db_definition = DatabaseDefinition(
+            ...    cloud_provider="aws",
+            ...    region="ap-south-1",
+            ...    capacity_units=4,
+            ...    keyspace="staging_xyz",
+            ...    pcu_group_id="01234567-89ab-cdef-0123-456789abcdef",
+            ... )
+            >>> asyncio.run(
+            ...     my_astra_db_admin.create_database(
```
erichare: Should this be async_create_database?

astrapy/admin/admin.py
```diff
@@ -1065,8 +1066,9 @@ def create_database(
         self,
         name: str,
         *,
-        cloud_provider: str,
-        region: str,
+        definition: DatabaseDefinition | None = None,
```
erichare, line 1099: My only question here is if we should use overloads https://typing.python.org/en/latest/spec/overload.html to basically allow either definition (but no cloud_provider or region), or cloud_provider and region (but no DatabaseDefinition. It's definitely nitpicky, but just looking at this function, at first glance and to a typechecker, it looks like i'm able to say create_database("my_name") with no additional arguments, which presumably would fail at runtime right?

Just an idea, not a blocker by any stretch.

erichare (approved):
LGTM @sl-at-ibm ! two small NITs but im approving

astrapy/admin/admin.py
sl-at-ibm: Verily! Well spotted :)

astrapy/admin/admin.py
sl-at-ibm: Good idea. For some reason I did not think of this. Let me rework the function definitions here.

sl-at-ibm: Done :>
Will push only after the CI for the other PR has finished ^_^

sl-at-ibm:
@skedwards88 FYI this is another, richer way to create a DB that e.g. allows to specify pcu stuff, nonvector etc. Noting that for an upcoming docs enrichment.

PR #422, opened 2026-06-17
chore: skip integration tests when no target DB is configured

erichare:
## Summary

Running **any** test in this repo currently requires a target Data API/DB to be configured. With no credentials set, `tests/conftest.py` imports `tests/preprocess_env`, which raises `ValueError("No credentials.")` at module-load time — aborting **collection of the entire suite, including the unit tests**:

```
ImportError while loading conftest 'tests/conftest.py'.
tests/preprocess_env.py:128: in <module>
    raise ValueError("No credentials.")
E   ValueError: No credentials.
```

This PR makes the **integration tests skip** (instead of crashing the run) when no target DB is configured, so contributors can run the unit suite with zero environment setup.

## What changed

- **`tests/preprocess_env.py`** — when none of `ASTRA_DB_API_ENDPOINT` / `LOCAL_DATA_API_ENDPOINT` / `DOCKER_COMPOSE_LOCAL_DATA_API` is set, record this via a new `TARGET_DB_EMPTY` flag (with safe defaults for `IS_ASTRA_DB`, `DOCKER_COMPOSE_LOCAL_DATA_API`, `SECONDARY_KEYSPACE`) instead of raising. The token-provider setup is guarded accordingly.
- **`tests/conftest.py`** — add a `pytest_collection_modifyitems` hook that, when `TARGET_DB_EMPTY`, skips every test located under an `integration/` directory **or** whose fixture closure needs a live database (`data_api_credentials_kwargs` & friends — `data_api_credentials_kwargs` is the only fixture that performs network calls during setup, and all integration tests depend on it transitively).
- **`CHANGES`**, **`DEVELOPING.md`** — documentation.

## Safety: no CI behavior change

When a target DB **is** configured (always the case in CI), `TARGET_DB_EMPTY` is `False` and the hook returns immediately — a complete no-op. `unit.yml` is intentionally left untouched (it still passes credentials), so the DB-dependent unit tests keep running there with full coverage.

## Test plan

Verified locally with **no** DB env configured:

- `pytest tests/base --collect-only` → **730 collected, no crash** (previously: hard `ValueError`).
- `pytest tests/base/unit` → **257 passed, 36 skipped, 0 DB errors**. *(One unrelated failure, `test_dataapitimestamp_lifecycle`, reproduces on `main` too — it's a local-timezone artifact and passes in UTC CI.)*
- `pytest tests/base/integration` → **436 skipped, 0 errors**.
- `pytest tests/admin tests/vectorize` → **19 skipped, 0 errors**.
- `TARGET_DB_EMPTY` is `True` with no creds, `False` with Astra creds (hook no-ops in CI).
- `ruff check tests`, `ruff format --check tests`, `mypy tests` → all clean.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -10,6 +10,7 @@ Bugfix: faulty normalization of empty/null indexing targets for `CollectionDefin
 maintenance: integration testing gets an embedding provider switcher logic
 maintenance: integration tests use the latest HCD (2.0.6) and a recent Data API build (1.0.46)
 maintenance: integration tests have configurable page size for find pagination
+maintenance: integration tests are skipped (instead of aborting collection) when no target DB is configured, so the unit-test suite can run without credentials
 
 v 2.2.1
 =======
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -65,7 +65,9 @@ Tests can be run on three types of Data API _targets_ (with slight differences i
 
 Depending on the target chosen, different environment variables are needed: refer to
 the `tests/env_templates/*.base.template` examples.
-Note that the variables defined in the desired "base" template **must** be set to run test, even for unit tests.
+Note that the variables defined in the desired "base" template **must** be set to run the integration tests.
+If no target DB is configured at all, the integration tests are automatically skipped, so the pure unit tests
+(those that need no database) can still be run with no environment setup.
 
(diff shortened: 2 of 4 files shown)


erichare:
Superseded by #423. The requirement was clarified to the opposite direction: the integration tests should be skipped when the target database is **non-empty** (to avoid running the destructive suite against a populated database), not when no DB is configured.

PR #423, opened 2026-06-17
chore: require an empty target database for integration tests

### LINKED ISSUE #419: Skip integration tests if target DB is non-empty
The base integration tests (tests/base/integration) are destructive, in that they freely create and drop collections, tables, and keyspaces on the configured target database. If the suite is accidentally pointed at a database that's already in use, it can clobber existing data. In addition, presence of target data is an indication of active CI ongoing.

We should guard against this: before running, check whether the working keyspace already contains any collections or tables. If it does, skip the entire integration suite rather than running (and mutating) a populated database.

**Proposed behavior**

On a target DB whose working keyspace is empty (no collections, no tables) → run the integration tests as normal.
On a non-empty target DB → skip the whole integration suite with a clear reason; do not create/drop anything.
The check must run before any test collection/table is provisioned, so a populated DB is never touched.

**Acceptance criteria**

Empty target keyspace = integration tests run unchanged (CI behavior unaffected, since CI starts from an empty DB).
Non-empty target keyspace = all integration tests reported as skipped, with a message identifying the offending collections/tables; no mutations occur.

erichare:
## Summary

Adds a fail-fast guard for integration tests when the target database is already populated.

The guard now:

- checks all non-system keyspaces, not just the configured working keyspace
- looks for collections, tables, and UDTs
- exits with an error locally and in CI so skipped local runs do not go unnoticed
- supports `TOLERATE_POPULATED_DATABASE=yes` for intentional narrow local test runs
- applies to both base and vectorize integration tests

This is meant to catch leftover test objects early, before a long run fails later with object-limit/conflict errors.

## Test plan

- `make format`
- Verified the guard fails before tests run when the target database contains existing objects
- Verified `TOLERATE_POPULATED_DATABASE=yes` bypasses the guard for intentional narrow local runs
- Verified the empty-database path still allows integration tests to start normally


sl-at-ibm:
Thanks for taking a stab at this feature!
However I have a few remarks that we should discuss before calling this PR final.

1. I would prefer the behaviour be the same between local and CI runs. When running tests locally,
whether targeting Astra or a local HCD, there are already several tests that are skipped for various reasons.
The risk here is for a test "skipped because DB populated" to go unnoticed when doing local development.

2. Also an important detail is that sometimes we run tests locally restricted to few, or just one, test function. In that case
we may want to tell the test setup "I know there's table xyz and collection abc already, but *trust me*, my test will fit nevertheless".
So I would request for an environment variable to silence this whole check:

```
TOLERATE_POPULATED_DATABASE="yes" uv run pytest tests/base/integration [...]
```

3. I'm OK with this feature covering only the 'base' portion of the tests (I'm not sure why the vectorize part should not get the
same safety treatment though. I understand that one is not part of CI but I'd say it wouldn't harm to make the caller conscious about
populated DBs.)

4. Unfortunately, I think, the checks need to be a bit wider than what is now in the fixture. The goal of this feature, the way I see it,
is not really to avoid destroying real data (we probably don't have a production collection called `id_test_collection` or so), rather
to avoid wasting 15+ minutes on a test run that ends up with a sad `TOO_MANY_COLLECTIONS` message which could have been prevented by exiting early.
In this view, we should ensure no collections are there _across all user keyspaces_. Adding a similar check for tables _and UDTs_ seems a good
addition (again, not to protect any "prod" data, but to avoid the occasional silly create-type or create-table fail 15+ minutes in
because of a leftover table that did not get cleaned up right from a previous test run).

So I would expand the check by ensuring there are no collections/UDTs/tables across all keyspaces that are not 'system keyspaces'.
This would require access to the database admin to list the keyspaces, and then a loop over keyspaces to collect existing stuff.
There is a detail here: on Astra, the Data API filters out the system keyspaces for us, whereas on HCD this must be done by our code.

Let me see how that could work (rought draft, I tested it in the basic cases):

```
from __future__ import annotations

import logging
import os

import pytest

from astrapy import Database

from .conftest import IS_ASTRA_DB

logger = logging.getLogger("cagnoli")


# this is effectively empty: the following system keyspaces are suppressed by the Data API itself on Astra
ASTRA_DB_SYSTEM_KEYSPACES: list[str] = [
    # "system_traces",
    # "datastax_sla"
    # "data_endpoint_auth",
    # "system_auth"
    # "system_views",
    # "system_schema"
    # "system_virtual_schema"
    # "system",
]
HCD_SYSTEM_KEYSPACES: list[str] = [
    "system_auth",
    "system_schema",
    "system_distributed",
    "system",
    "system_traces",
    "system_views",
    "system_virtual_schema",
]


@pytest.fixture(scope="module", autouse=True)
def require_empty_target_database(sync_database: Database) -> None:
    if os.environ.get("TOLERATE_POPULATED_DATABASE", "").strip() != "":
        return

    collected_items: dict[str, list[str]] = {
        "collections": [],
        "tables": [],
        "types": [],
    }
    system_keyspaces = (
        ASTRA_DB_SYSTEM_KEYSPACES if IS_ASTRA_DB else HCD_SYSTEM_KEYSPACES
    )
    keyspaces = [
        ks
        for ks in sync_database.get_database_admin().list_keyspaces()
        if ks not in system_keyspaces
    ]
    for keyspace in keyspaces:
        _db = sync_database._copy(keyspace=keyspace)
        for collection in _db.list_collection_names():
            collected_items["collections"].append(f"{keyspace}.{collection}")
        for table in _db.list_table_names():
            collected_items["tables"].append(f"{keyspace}.{table}")
        for type_ in _db.list_type_names():
            collected_items["types"].append(f"{keyspace}.{type_}")
    if any(itm_list != [] for itm_list in collected_items.values()):
        items_summary = "; ".join(
            f"{kind}: {', '.join(sorted(itm_list))}"
            for kind, itm_list in sorted(collected_items.items())
            if itm_list != []
        )
        message = f"Non-empty target database detected. Items found: {items_summary}."
        pytest.exit(message, returncode=1)

@pytest.mark.describe("Test cagnoli")
def test_cagnoli() -> None:
    pass
```

WDYT? Keep in mind this can be de-prioritized if we are tight on the 2.3-release things! :)

erichare:
Thank you @sl-at-ibm :) I agree with you. I updated the fixture so the behavior is now consistent locally and in CI: it exits early with an error if the target database is populated, unless `TOLERATE_POPULATED_DATABASE=yes` is set.

i also applied it across keyspaces... it filters system keyspaces, and checks collections/tables/UDTs across all non-system keyspaces. I used `with_options(keyspace=...)` rather than `_copy`, but otherwise tried to stay faithful to your example!

I left this scoped to the base integration suite for now since that is the always-run/CI-covered suite, but I agree the same guard could be extracted/shared for vectorize tests.

In fact, heck, let's add that really quick. I was going to say it could go in a follow up, but it probably makes sense just to have it here!... Done.

By the way, the i hardcoded the list of system keyspaces... do you know of a simple data-api way to get that list dynamically? i can update it if so...

sl-at-ibm:
Awesome, Eric, thank you!

> By the way, the i hardcoded the list of system keyspaces... do you know of a simple data-api way to get that list dynamically? i can update it if so...

I am not aware of such a handy cool thing and I highly doubt there is one. For now, at least, we can safely keep the hardcoded list without worrying too much (it's not even user-facing stuff anyway ...)

tests/empty_database_guard.py
```diff
@@ -0,0 +1,131 @@
...
+
+import pytest
+
+from astrapy import Database
+
+TOLERATE_POPULATED_DATABASE_ENV = "TOLERATE_POPULATED_DATABASE"
+
+ASTRA_DB_SYSTEM_KEYSPACES = frozenset(
+    {
+        "data_endpoint_auth",
```
sl-at-ibm: No real need to have this set nonempty since for Astra the Data API filters them out already - but it doesn't hurt to make the list explicit I suppose.

tests/empty_database_guard.py
```diff
@@ -0,0 +1,131 @@
...
+        "system_schema",
+        "system_traces",
+        "system_views",
+        "system_virtual_schema",
+    }
+)
+
+
+def _is_populated_database_tolerated() -> bool:
+    return os.environ.get(TOLERATE_POPULATED_DATABASE_ENV, "").strip().lower() in {
```
sl-at-ibm: Suggestion to use `extended_booleanize_env` in `preprocess_env.py` rather than a (somewhat fragile) hardcoded list - right now, "3", "-1" or "Yes" wouldn't count as true.

(full disclosure: I had forgotten about that handy utility for at least one year, now I re-discovered it lol)

sl-at-ibm (approved):
LGTM, thanks for doing that! (you may want to check the suggestion I left)

tests/empty_database_guard.py
erichare: Good call! updated

tests/empty_database_guard.py
erichare: Holy moly i didn't know that!! Or if i did, i forgot. That's awesome haha. Just updated to use it

sl-at-ibm (approved):

Issue #29, opened 2023-10-23
Makefile/linter/codestyle support


hemidactylus:
Something like a `ruff/black/mypy` suite as an easy-to-use make command would encourage good writing style IMO

hemidactylus:
I suggest a Makefile would still one day be useful to have mypy+linter and possibly a nifty "make test" command easily accessible. @erichare possibly a new issue with that wdyt?

Issue #27, opened 2023-10-23
Raw response return in utils / make_requests


hemidactylus:
In:
```
    try:
        if return_type == "json":
            return r.json()
        else:
            return r
```
shouldn't it be ` r.text` at the end?

erichare:
@hemidactylus the reason for this.... is for reasons that i'm not quite sure of. It was in the existing package and i just left it. For create and terminate database, we return:

```
    def create_database(self, database_definition=None):
        r = self._ops_request(
            method=http_methods.POST,
            path="/databases",
            json_data=database_definition,
            return_type="response"
        )

        if r.status_code == 201:
            return {"id": r.headers["Location"]}

        return None

    def terminate_database(self, database=""):
        r = self._ops_request(
            method=http_methods.POST,
            path=f"/databases/{database}/terminate",
            return_type="response"
        )

        if r.status_code == 202:
            return database

        return None
```

I'm wondering if actually we just always want to return the response object, so any downstream function can process it, the status code, etc as needed

erichare:
@hemidactylus - i've cleaned this up. I think it's best that we return the raw response. The reason being is that in theory each api call can have different behavior depending on the status code, etc. Right now we don't carefully check that very often, but we can and should. So that utility function will return the raw response - we can extract the json data, the text, the status_code, and any other property of it. I think that'll keep things more extensible.

Issue #26, opened 2023-10-23
Protect against empty "filter" clause not accepted by the API


hemidactylus:
Currently the vector API breaks on find if empty dicts are passed for example for sort.
  In other words, `..."sort": {}...` in the payload causes errors, while omitting it works all right.
  While this will be addressed, I suggest to clean out empty dicts from such places before issuing requests. I don't think it would hurt and it would get all the stack to work sooner.

Issue #23, opened 2023-10-23
logging of all APi requests/responses


hemidactylus:
optional, maybe with a wrapper class or Logger.

Opt=in to log the vector values, a mechanism (on by default) to purge them from outputs for readability

PR #424, opened 2026-06-17
chore: bump idna and pygments in uv.lock to clear Dependabot alerts

erichare:
## Summary

Resolves the Dependabot security alerts that can be cleared **without dropping Python 3.9 support** — a lock-file-only change (transitive dependencies, no `pyproject.toml` constraints touched):

| Alert | Package | Bump | Advisory | Patched in | py3.9? |
|------|---------|------|----------|-----------|--------|
| #32 (medium) | `idna` | 3.10 → 3.18 | `idna.encode()` can bypass the CVE-2024-3651 fix | 3.15 | ✅ requires `>=3.8` |
| #27 (low) | `pygments` | 2.19.2 → 2.20.0 | ReDoS in GUID-matching regex | 2.20.0 | ✅ requires `>=3.9` |

Both patched versions still support Python 3.9, so uv resolves each to a single (non-forked) version and the alerts clear.

## Why the other 7 alerts are not included

The remaining open alerts — `urllib3` (#30, #31), `requests` (#26), `python-dotenv` (#29), `pytest` (#28), `filelock` (#20, #23) — **cannot be cleared while the project supports Python 3.9**. Every one of their first-patched versions requires `python >=3.10`:

- `urllib3 2.7.0`, `requests 2.33.0`, `python-dotenv 1.2.2`, `pytest 9.0.3`, `filelock 3.20.3` → all `requires-python >=3.10`.

With `requires-python = ">=3.9"`, uv forks the resolution: it installs the patched version for py3.10+ but **keeps the vulnerable version pinned for the py3.9 branch** of `uv.lock`. Dependabot scans the whole lock, so the alert stays open. (`filelock` already demonstrates this today — it's at the safe 3.24.3 for py3.10+ but stuck at 3.18.0 for py3.9.) These are the same exceptions documented in #398 ("address dependabot vulns except incompatible with supporting py 3.9").

Clearing those 7 requires a separate decision to drop Python 3.9 (already EOL as of Oct 2025), which is out of scope for a routine version-bump PR.

## Test plan

- [x] `ruff check astrapy tests` — passes
- [x] `ruff format --check astrapy tests` — passes
- [x] `mypy astrapy tests` — passes (215 files, no issues)
- [x] `pytest tests/base/unit` — no new failures vs. baseline (idna/pygments are transitive deps unused at the astrapy layer)
- [ ] CI green

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -10,6 +10,7 @@ Bugfix: faulty normalization of empty/null indexing targets for `CollectionDefin
 maintenance: integration testing gets an embedding provider switcher logic
 maintenance: integration tests use the latest HCD (2.0.6) and a recent Data API build (1.0.46)
 maintenance: integration tests have configurable page size for find pagination
+maintenance: bump idna (3.10->3.18) and pygments (2.19.2->2.20.0) in uv.lock to clear Dependabot alerts (remaining alerts require dropping Python 3.9)
 
 v 2.2.1
 =======
--- a/uv.lock
+++ b/uv.lock
@@ -582,11 +582,11 @@ wheels = [
 
 [[package]]
 name = "idna"
-version = "3.10"
+version = "3.18"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/f1/70/7703c29685631f5a7590aa73f1f1d3fa9a380e654b86af429e0934a32f7d/idna-3.10.tar.gz", hash = "sha256:12f65c9b470abda6dc35cf8e63cc574b1c52b11df2c86030af0ac09b01b13ea9", size = 190490, upload-time = "2024-09-15T18:07:39.745Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/cd/63/9496c57188a2ee585e0f1db071d75089a11e98aa86eb99d9d7618fc1edce/idna-3.18.tar.gz", hash = "sha256:ffb385a7e039654cef1ab9ef32c6fafe283c0c0467bba1d9029738ce4a14a848", size = 196711, upload-time = "2026-06-02T14:34:07.794Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/76/c6/c88e154df9c4e1a2a66ccf0005a88dfb2650c1dffb6f5ce603dfbd452ce3/idna-3.10-py3-none-any.whl", hash = "sha256:946d195a0d259cbba61165e88e65941f16e9b36ea6ddb97f00452bae8b1287d3", size = 70442, upload-time = "2024-09-15T18:07:37.964Z" },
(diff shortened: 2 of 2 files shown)


sl-at-ibm:
You know... I start to think nobody would get mad at astrapy for dropping a long-EOL'd version of Python. Let's get this one in, then we update min-python-required, so that the other CVEs can be addressed (filelock and the others).

- This will mean a couple of adjustments here and there (there's a "min python version" somewhere in the ci/cd, plus pyproject, readme, etc).
- But I was mistaken: the next language simplification comes when 3.10 will be dropped (it's the way `Self` is imported in context_managers.py)
- On the other hand, 3.9+ would make it possible to restore using a modern testcontainers, which right now is partially replaced/subclasses/monkeypatched because ...um... (checks code) invoke `docker compose` rather than `docker-compose` (which would not run in the GH runners). <-- _Definitely not urgent lol_

sl-at-ibm:
Generally all ok, but just a question for my own education: these are changes to the lockfile only (as opposed to pyproject explicit pinning).
What happens when I next run `uv lock` as part of some other work and then commit: wouldn't thes present changes be potentially lost/overwritten?

(I am interested in the answer because in the past I have [addressed some earlier CVEs with pyproject-explicit pinning](https://github.com/datastax/astrapy/pull/398/changes) (of the `>=` kind of course) and would like to learn if that was too much/if it can be made less intrusive as a post-edit).

erichare:
> Generally all ok, but just a question for my own education: these are changes to the lockfile only (as opposed to pyproject explicit pinning). What happens when I next run `uv lock` as part of some other work and then commit: wouldn't thes present changes be potentially lost/overwritten?
> 
> (I am interested in the answer because in the past I have [addressed some earlier CVEs with pyproject-explicit pinning](https://github.com/datastax/astrapy/pull/398/changes) (of the `>=` kind of course) and would like to learn if that was too much/if it can be made less intrusive as a post-edit).

@sl-at-ibm I recently had a task like this in Langflow and had the same question. My understanding is:

`pyproject.toml` is where we declare project-level constraints, so it makes sense to edit it for direct astrapy dependencies. But these two are transitive dependencies (i think idna from httpx and pygments from ipython), so adding them there would promote them into explicit project dependencies just to influence the resolver.

For `uv.lock`, plain `uv lock` uses the existing lockfile as preferences. So after this is merged, a future ordinary `uv lock` should preserve these resolved versions. They could still change if someone runs `uv lock --upgrade` / `--upgrade-package`, or if a future direct dependency change imposes different constraints.

If a future direct dependency actually forced an old vulnerable version, I think that would be a deeper dependency problem to address. Adding the transitive package directly to `pyproject.toml` probably would not solve that cleanly; it would more likely create a resolver conflict. hopefully that wouldn't happen haha

That was longer than i intended haha but i had to convince myself as well :) Does that make sense?

sl-at-ibm (approved):
Thank you for the explanation! I think it makes sense, yes.
(A couple of times it has happened to me on some PRs that I found myself with a botched lockfile, so I deleted it and ran a fresh `uv lock`. This may potentially be trouble, I don't know -- perhaps the right thing in that case is to restore the lockfile from main and run "ordinary uv lock" as you mention.)

Anyway, there's no big risk in going forward like this PR does and keep an eye if some weird dynamics brings lower vulnerable versions around again (unlikely as it may be...)

So: LGTM!

PR #425, opened 2026-06-18
Full adaptation of CI/CD to parametrized FIND_PAGE_SIZE

sl-at-ibm:
In preparation to deploy of Data API v1.0.48, with (deploy-time) configurable page size, this PR completes the adaptation of integration tests to cope with this change.

The changes here are inert unless the tests are run after e.g. `export FIND_PAGE_SIZE=50`. So this can be merged any time now. Once Astra prod gets the new page size (of 50 by default), setting a new repo env variable will make the tests run smoothly.
At that point, it might be even simpler to bump the HCD compose data api version to 1.0.48 and hardcode 50 again.

Note that right now the workflow yamls transmit the repo secret (for `FIND_PAGE_SIZE` all the way to the _Astra_ integration test only, because this repo has full control over the HCD case (and will manage the version used there)).


tests/base/integration/collections/test_collection_cursor_async.py
```diff
@@ -26,9 +27,10 @@
 
 from ..conftest import DefaultAsyncCollection
 
-PAGE_SIZE = 20  # TODO: set to 50, default as per Data API config after PR 2461
-NUM_DOCS = 2 * PAGE_SIZE + 5
-NUM_DOCS_PAGINATION = 2 * (2 * PAGE_SIZE) + 5
+# TODO: once v1.0.48 is in production, hardcode 50 and bump Data API version in docker compose file:
+FIND_PAGE_SIZE = int(os.environ.get("FIND_PAGE_SIZE") or "20")
```
erichare, line 31: Do we want the fallback value to be 20 here, or should it be 50 to match the new server default? 

Oh, LOL i typed that before i read your comment above. Makes sense we should wait till its in production!

DEVELOPING.md
```diff
@@ -171,6 +171,15 @@ certain environment variables, otherwise the associated tests are excluded from
 Prepend tests with a `ASTRAPY_TEST_LATEST_MAIN=y` for features found on `main` that are not released anywhere.
 _(Tip: run a code search first to see what is currently marked as such. Chances are nothing is.)_
 
+### Page size increase (starting in v1.0.48)
```
erichare, line 174: Perfect. I need to read the whole PR before commenting in the future 😄

tests/base/integration/collections/test_collection_cursor_sync.py
erichare, line 1: Do we need to make the corresponding changes in test_table_cursor_sync as well?

.github/workflows/main.yml
```diff
@@ -18,6 +18,10 @@ on:
         required: true
       HEADER_EMBEDDING_API_KEY_VOYAGEAI:
         required: true
+      FIND_PAGE_SIZE:
```
erichare, line 21: Do we also need to include FIND_PAGE_SIZE in the release workflow? i think it runs the integration tests directly

erichare (changes requested):
Left a couple comments @sl-at-ibm, possible places that might also need adapting (not 100% sure) hence the request changes. Let me know if im off base though and ill approve!

tests/base/integration/collections/test_collection_cursor_sync.py
sl-at-ibm: Very well spotted, thank you! I am adding that as well right now.

.github/workflows/main.yml
sl-at-ibm: Once more, you are perfectly right! Added.

erichare (approved):
LGTM @sl-at-ibm !

PR #427, opened 2026-06-23
Issue 426: List PCU Groups and its impact on create_database

### LINKED ISSUE #426: AstraDBAdmin: add PCU Group listing capabilities
As a companion feature to the improved `create_database` (issue #421), this is about the AstraDBAdmin allowing listing of "PCU Groups" for the current org.

This is required because:

1. the logical process by a PCU-using user is to first query the PCU Groups, so that they can select one and retrieve its UUID to supply to `create_database`
2. also, as per client design, if a pcu group is supplied to create_database, the client will need to ensure it exists for the right provider/region. (this is an exception to the general rule of clients not interfering with checks and validation, leaving it to the server-side; but in this case it is motivated by the lack of server checks and the huge advantage in having the operation fail immediately, rather than leaving a DB in an inconsistent/undefined state, broken in mid-creation).

## Specs

Three things: validation of pcu group id if passed to create_database + dataclasses to describe pcu groups + method to retrieve pcu groups.

### Validation

If a non-null `pcu_group_id` is passed in the `definition` parameter to create_database (see PR #421 ), a pre-check on the client side is done to ensure the PCU Group exists. If not, a DevOpsAPIException is raised right away and no create-db request is fired.

### Listing PCU Groups

`AstraDBAdmin` must offer a method (2x, sync/async) to list the PCU Groups for the current org, following the corresponding `POST /v2/pcu/actions/get` [DevOps API endpoint](https://docs.datastax.com/en/astra-api-docs/_attachments/devops-api/index.html#tag/PCU/operation/pcuGet).

According to the yellow doc, clients may add a filtering (in which case both `cloud_provider` and `region` are required): so,

- `list_pcu_groups(self) -> list[PCUGroupDescriptor]`
- `list_pcu_groups(self, *, str cloud_provider, str region) -> list[PCUGroupDescriptor]`
- `async_list_pcu_groups(self) -> list[PCUGroupDescriptor]`
- `async_list_pcu_groups(self, *, str cloud_provider, str region) -> list[PCUGroupDescriptor]`

### Descriptor

The `PCUGroupDescriptor` new dataclass must mimic closely the response from the DevOps API, with:
- UUIDs (coming as strings) left strings
- datetimes (as string) parsed into datetimes
- no enums (for future-flexibility), just strings there
- a nested dictionary translated into a nested object (see below)

Also the standard rename conventions take place:
- `id` in place of the response's `uuid`
- all other fields trivially snake-case versions of the payload counterpart (with all-uppercase `PCU` in class name and all-lowercase `pcu`, `id` and such in the member names)

Structure:

```
PCUGroupDescriptor:
    id: str
    org_id: str
    title: str
    cloud_provider: str
    region: str
    instance_type: str
    pcu_type: PCUGroupTypeDescriptor
    provision_type: str
    min: int
    max: int
    reserved: int
    description: str
    created_at: datetime
    updated_at: datetime
    created_by: str
    updated_by: str
    status: str

PCUGroupTypeDescriptor:
    type: str
    region: str
    cloud_provider: str (NOTE: 'provider' in payload)
    details: PCUGroupTypeDetailsDescriptor

PCUGroupTypeDetailsDescriptor:
    v_cpu: int
    memory: str
    disk_cache: string
```

sl-at-ibm:
Fixes #426 

1. dataclasses to describe the PCU Groups
2. AstraDBAdmin methods to list the PCU Groups for an org (incl. client-side provider+region filtering; also a/sync)
3. pre-check for the validity of PCU Group ID (if one is passed) during create_database admin method
4. reinstated handling of the PCU aspects of the returned regions for `find_available_regions` admin method(s)
5. all relevant tests (unit tests with a local httpserver pretending to be the devops api)

Code-complete as far as I can see. The details on naming conventions are still awaiting a Client Team formal approval, but no other changes should be on their way (i.e. this could also be merged as is and be refined later if needed - probably it's ok as is)


astrapy/data/info/database_info.py
```diff
@@ -350,6 +350,7 @@ class AstraDBAvailableRegionInfo:
     name: str
     reserved_for_qualified_users: bool
     zone: str
+    pcu_types: list[PCUGroupTypeDescriptor] | None
```
erichare: Should `pcu_types` default to None @sl-at-ibm ? For people that were calling this function prior to this change?

tests/base/unit/test_admin_create_database_sync.py
```diff
@@ -158,19 +228,118 @@ def test_admin_create_database_definition_sync(
...
+        )
+
+        db_definition = DatabaseDefinition(
+            cloud_provider="cp",
+            region="r",
+            keyspace="ks",
+            tier="tr",
+            capacity_units=5,
+            db_type="ty",
+            pcu_group_id="d",
```
erichare: To test "matched cpu", should we be providing this dynamically from the above calls, instead of using `"d"`? Same comment for the async version

astrapy/admin/admin.py
```diff
@@ -2273,6 +2341,200 @@ async def async_find_available_regions(
...
+        logger.info("getting PCU groups (DevOps API)")
+        # this response can be in fact a JSON list or even a 'null':
+        lp_response = cast(
+            list[dict[str, Any]],
+            self._orgwide_dev_ops_api_commander.request(
+                http_method=HttpMethod.POST,
+                payload={},
+                additional_path="pcus/actions/get",
+                timeout_context=timeout_ctx,
+                caller_function_name="find_available_regions",
```
erichare: I think this should be `"list_pcu_groups"` right?

astrapy/admin/admin.py
```diff
@@ -2273,6 +2341,200 @@ async def async_find_available_regions(
...
+        logger.info("getting PCU groups (DevOps API)")
+        # this response can be in fact a JSON list or even a 'null':
+        lp_response = cast(
+            list[dict[str, Any]],
+            await self._orgwide_dev_ops_api_commander.async_request(
+                http_method=HttpMethod.POST,
+                payload={},
+                additional_path="pcus/actions/get",
+                timeout_context=timeout_ctx,
+                caller_function_name="find_available_regions",
```
erichare: `"async_list_pcu_groups"` here?

erichare (changes requested):
Left a few small comments @sl-at-ibm ! Let me know if you agree. Very nice PR :)

astrapy/data/info/database_info.py
sl-at-ibm: Good point. I presume zero users are interested in creating instances of `AstraDBAvailableRegionInfo` other than in the `find_available_regions` method ... but it does not hurt to enforce formal backward-compatibility.
I appreciate the rigour :+1: 
(added).

tests/base/unit/test_admin_create_database_sync.py
sl-at-ibm: You are very right! I am adapting it but I have to leave for a couple of hours ... so will finish later today! Awesome and thank you for the review!

astrapy/data/info/database_info.py
erichare: yep hahaha. It's probably backwards compatibility that exactly no one will benefit from, but if there's no consequence to set it that way, then I guess its okay! Only counter argument is that all the other parameters are required, so it stands out a bit haha

tests/base/unit/test_admin_create_database_sync.py
sl-at-ibm: done.

erichare (approved):
LGTM @sl-at-ibm !

PR #428, opened 2026-06-25
Minor cursor behaviour improvements (+ testing)

sl-at-ibm:
This PR fixes/enhances a couple of details in the cursor testing and introduces two slight changes in the cursor behaviour, related to reflecting more faithfully the cursor state (IDLE -> STARTED -> CLOSED) under certain operations.

_(it also denylists some unreleased vectorize models to be able to run vectorize tests on certain dev deployments, but this is irrelevant)_

The cursor behaviour changes, dictated by a better alignment with the Typescript client cursor behaviour, are so tiny that I do not think they should be treated as "breaking changes", rather as essentially bug fixes. "started" means we started issuing request, and "closed" means there is nothing more to fetch, so...

### DIFF
--- a/astrapy/data/cursors/find_cursor.py
+++ b/astrapy/data/cursors/find_cursor.py
@@ -1866,7 +1866,7 @@ class TableFindCursor(Generic[TRAW, T], AbstractCursor[TRAW]):
         This operation is allowed only if the cursor state is still IDLE.
 
         Instead of explicitly invoking this method, the typical usage consists
-        in passing arguments to the Collection `find` method.
+        in passing arguments to the Table `find` method.
 
         Args:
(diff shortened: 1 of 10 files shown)


astrapy/data/cursors/find_cursor.py
erichare: Should this comment be updated now that STARTED is set?

erichare (approved):
LGTM @sl-at-ibm, one NIT comment

astrapy/data/cursors/find_cursor.py
sl-at-ibm: Ah. That's correct. Thanks for spotting this!

Fixed. (6 separate docstrings).

PR #430, opened 2026-06-26
Remove deprecated region_name property alias

sl-at-ibm:
That was deprecated in 2.0.1 and 2.3 is the time for removal from `AstraDBAvailableRegionInfo` and `AstraDBAdminDatabaseRegionInfo`.


erichare (approved):
LGTM

PR #431, opened 2026-06-28
Minor adjustments pre-v2.3

sl-at-ibm:
- (following Cédrick) Different errors if the PCU Group ID is outright nonexistent or belongs to a different cloud provider/region (+ specific tests)
- the error raised is not ValueError anymore, rather DevOpsAPIException
- more robust unmarshaling of pcu-related devops api responses (= everything that is not critical is made optional, no errors if fields missing from responses)

### DIFF
--- a/astrapy/admin/admin.py
+++ b/astrapy/admin/admin.py
@@ -1256,7 +1256,7 @@ class AstraDBAdmin:
                     "PCU Group ID pre-check failed. Aborting database creation."
                 )
                 if matching_pcu_groups == []:
-                    raise ValueError(
+                    raise DevOpsAPIException(
                         f"Requested PCU Group ID '{_definition.pcu_group_id}' not found for cloud provider "
                         f"('{_definition.cloud_provider}') and region ('{_definition.region}')."
                     )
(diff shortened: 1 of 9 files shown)


astrapy/admin/admin.py
```diff
@@ -1236,30 +1236,50 @@ def create_database(
...
+                    )
+                    raise DevOpsAPIException(
+                        f"Requested PCU Group ID '{_definition.pcu_group_id}' not "
+                        "found for cloud provider provider/region "
+                        f"('{_definition.cloud_provider}' / '{_definition.region}'). "
+                        "Aborting database creation."
+                    )
+                else:
+                    # is the matching group in the right cloud provider / region?
+                    expected_cpr = (_definition.cloud_provider, _definition.region)
```
erichare: Going back to the whole "aws" vs "AWS" discussion... should this perhaps be a case-insensitive comparison? Or do we want to enforce a strict case sensitive match

erichare (approved):
LGTM @sl-at-ibm , one small question regarding case sensitivity but looks great otherwise

astrapy/admin/admin.py
```diff
@@ -1513,32 +1533,52 @@ async def async_create_database(
...
+                    )
+                    raise DevOpsAPIException(
+                        f"Requested PCU Group ID '{_definition.pcu_group_id}' "
+                        "not found for cloud provider provider/region "
+                        f"('{_definition.cloud_provider}' / '{_definition.region}'). "
+                        "Aborting database creation."
+                    )
+                else:
+                    # is the matching group in the right cloud provider / region?
+                    expected_cpr = (_definition.cloud_provider, _definition.region)
```
erichare: Same question as up above, basically if this side of the comparison should also be case-normalized as well

sl-at-ibm: Thanks for spotting those! Fixed.

I will also bundle the bump to 2.3 in this PR and then proceed with releasing. (to lessen the business with getting the CI go green).

PR #432, opened 2026-07-08
Support nonvector DB creation

sl-at-ibm:
by means of a client-only `db_type = "nonvector"` choice.

This was the agreement in the client team to overcome the expected payloads from the DevOps API.

- Tested manually creating actual DBs in various cases (admin tests are not covered by ci/cd);
- added extensive unit test against a local mock http server.


erichare (approved):
LGTM @sl-at-ibm !

PR #434, opened 2026-08-19
Update README to point to HCD 2.0

skedwards88:
Update the README to point to the HCD 2.0 docs instead of the 1.0 docs


erichare (approved):
Looks good to me! and also seems that https://docs.datastax.com/en/hyper-converged-database/2.0/connect/python-client.html has a redirect, which is nice!

PR #435, opened 2026-09-21
Maintenance patching

sl-at-ibm:
- Bump data api version for HCD testing
- Bump some transitive dependencies for remediating CVEs (anyio, setuptools, h2)
- Split table/collection page-size env var management; complete handling of these settings in the CI workflows
- switch to the preferred `"nonvector"` form for the non-vector special DB type setting in `DatabaseDefinition`

Closes https://github.com/datastax/astrapy/security/dependabot/36
Closes https://github.com/datastax/astrapy/security/dependabot/35
Closes https://github.com/datastax/astrapy/security/dependabot/34
Closes https://github.com/datastax/astrapy/security/dependabot/33


sl-at-ibm:
Force-merging because the only blocker is the fact that in computing the testing coverage the artifact from main is expired. (will raise its retention period from 30 to 90 days, the max allowed).

